PR #4, opened 2020-12-26
Reservation routes

florisdobber:


service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private final DSLContext db;
+
+    public ReservationProcessorImpl(DSLContext db) {
+        this.db = db;
+    }
+
+
+    @Override
+    public void makeReservation(JWTData userData, MakeReservationRequest makeReservationRequest) {
```
liammoyn, line 24: Needs checks for
1. If the userId exists
2. If the user has the privilege level of executing this action (May not be relevant)
2. If the blockId exists
3. If the teamId exists
4. If the given user is on the given team

A lot of these checks will be repeated for all of these routes so feel free to try to pull that into a helper function if it comes out cleanly.

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,206 @@
...
+  public Optional<ReservationsRecord> lastAction(int blockId) {
+    return Optional.ofNullable(
+        db.selectFrom(RESERVATIONS)
+            .where(RESERVATIONS.BLOCK_ID.eq(blockId))
+            .orderBy(RESERVATIONS.PERFORMED_AT.desc())
+            .limit(1)
+            .fetchOne());
+  }
+
+  public void basicChecks(int blockId, Integer userId, Integer teamId) {
```
liammoyn, line 32: You can use the fetchExists clause in jooq to directly do these checks. The only case where you'd want to actually fetch the record is if you then want to run some operation on it but it doesn't look like you're doing that here

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,206 @@
...
+
+    Optional<ReservationsRecord> maybeReservation = lastAction(reservationsRecord.getBlockId());
+
+    // check if the user is an admin and if the block has been completed
+    if (userData.getPrivilegeLevel().equals(PrivilegeLevel.STANDARD)) {
+      throw new AuthException("User does not have the required privilege level.");
+    }
+
+    if (maybeReservation.isPresent()) {
+      if (!(maybeReservation.get().getActionType().equals("complete"))) {
```
liammoyn, line 169: Why are we comparing strings instead of enums here?

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,206 @@
...
+    reservationsRecord.setBlockId(makeReservationRequest.getBlockID());
+    reservationsRecord.setUserId(userData.getUserId());
+    reservationsRecord.setTeamId(makeReservationRequest.getTeamID());
+    reservationsRecord.setActionType("reserve");
+    reservationsRecord.setPerformedAt(new Timestamp(System.currentTimeMillis()));
+
+    Optional<ReservationsRecord> maybeReservation = lastAction(reservationsRecord.getBlockId());
+
+    // check if block is open (released, uncomplete or no record)
+    if (maybeReservation.isPresent()) {
```
liammoyn, line 85: I don't love how much logic is stored in these if statements at the end of each of these functions. Can you pull them out into helper functions that have javadoc that explain the conditions for when they throw exceptions? These are the places that are most likely to change if we change any reservation logic so it'll be good to have them already separated and written as clearly as possible

liammoyn (changes requested):

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
florisdobber: Enum wasn't behaving as expected and forgot to fix, but now resolved

api/src/main/java/com/codeforcommunity/rest/FailureHandler.java
```diff
@@ -1,17 +1,6 @@
...
-import com.codeforcommunity.exceptions.HandledException;
-import com.codeforcommunity.exceptions.InvalidSecretKeyException;
-import com.codeforcommunity.exceptions.MalformedParameterException;
-import com.codeforcommunity.exceptions.MissingHeaderException;
-import com.codeforcommunity.exceptions.MissingParameterException;
-import com.codeforcommunity.exceptions.TokenInvalidException;
-import com.codeforcommunity.exceptions.UsedSecretKeyException;
-import com.codeforcommunity.exceptions.UserDoesNotExistException;
-import com.codeforcommunity.exceptions.UsernameAlreadyInUseException;
+import com.codeforcommunity.exceptions.*;
```
liammoyn, line 3: This isn't a huge deal but I'd suggest never using wildcard imports. You can change the setting on Intellij so that it doesn't do this automatically.
Wildcards aren't great because it can make it harder to tell where certain code is coming from. And because 90% of the time imports are going to be collapsed it doesn't really matter how long our import list is.

florisdobber: Yeah that was done automatically, I'll fix it in a separate PR, I broke this one

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,233 @@
...
+
+    if (teamId != null) {
+      if (!db.fetchExists(db.selectFrom(TEAMS).where(TEAMS.ID.eq(teamId)))) {
+        throw new ResourceDoesNotExistException(teamId, "team");
+      }
+    }
+
+    if (userId != null && teamId != null) {
+      if (!db.fetchExists(
+          db.selectFrom(USERS_TEAMS)
```
liammoyn, line 52: Do we still have the idea of pending user_team requests? If so, we need to exclude them from this db query

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,233 @@
...
+      if (maybeReservation.get().getUserId().equals(userId)) {
+        return;
+      }
+    }
+
+    // check if a team the user is on reserved the block, if they did, return
+    if (maybeReservation.get().getTeamId() != null) {
+      int teamId = maybeReservation.get().getTeamId();
+      if (db.fetchExists(
+          db.selectFrom(USERS_TEAMS)
```
liammoyn, line 118: Would also have to exclude pending user_team entries here

liammoyn (commented):
Right now if a block is marked for QA there's no way for a user to change its status, what's our desired flow for blocks after being marked QA?

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+    if (userId != null && teamId != null) {
+      if (!db.fetchExists(
+          db.selectFrom(USERS_TEAMS)
+              .where(USERS_TEAMS.USER_ID.eq(userId))
+              .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+              .and(
+                  USERS_TEAMS
+                      .TEAM_ROLE
+                      .eq(TeamRole.MEMBER)
```
liammoyn, line 59: I would prefer if we changed this to checking if TEAM_ROLE is **not** PENDING. This is because pending is kind of a special case and it's much more likely we might add another valid team role (like admin or super member, ext...) than it would be to add a role that behaves like pending. That means we would have to know to change all these calls as well which is an easy way to get us in trouble.

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    if (maybeReservation.get().getTeamId() != null) {
+      int teamId = maybeReservation.get().getTeamId();
+      if (db.fetchExists(
+          db.selectFrom(USERS_TEAMS)
+              .where(USERS_TEAMS.USER_ID.eq(userId))
+              .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+              .and(
+                  USERS_TEAMS
+                      .TEAM_ROLE
+                      .eq(TeamRole.MEMBER)
```
liammoyn, line 130: Same thing here, change to not PENDING. (Or use that helper function :))

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
liammoyn: We could also factor this clause into it's own helper function somewhere: `isOnTeam(int userId, int teamId)` so that we only have a single source of truth for this clause (which is always preferable)

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,300 @@
...
+            .limit(1)
+            .fetchOne());
+  }
+
+  private void basicChecks(int blockId, Integer userId, Integer teamId) {
+    if (!db.fetchExists(db.selectFrom(BLOCKS).where(BLOCKS.ID.eq(blockId)))) {
+      throw new ResourceDoesNotExistException(blockId, "block");
+    }
+
+    if (userId != null) {
```
liammoyn, line 39: This is nitpicky but all of the double nested if clauses you have in this file can be combined into a single if. Java has boolean shortcircuiting so if userId is equal to null then the second clause won't execute so there's no chance of a Null Pointer Exception

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
liammoyn: Just read the doc and realized you also have to include **not** NONE but I still think reversing the phrasing is worth it

florisdobber: This change has been made

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
florisdobber: Did most of these

api/src/main/java/com/codeforcommunity/dto/reservation/FailQARequest.java
```diff
@@ -0,0 +1,36 @@
+package com.codeforcommunity.dto.reservation;
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FailQARequest extends ApiDto {
```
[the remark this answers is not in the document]
florisdobber: Done

api/src/main/java/com/codeforcommunity/exceptions/IncorrectBlockStatusException.java
```diff
@@ -0,0 +1,28 @@
+package com.codeforcommunity.exceptions;
+
+import com.codeforcommunity.rest.FailureHandler;
+import io.vertx.ext.web.RoutingContext;
+
+public class IncorrectBlockStatusException extends HandledException {
+
+  private final int blockId;
+  private final String expectedStatus;
```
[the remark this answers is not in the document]
florisdobber: Not sure how I would go about this so not doing this.

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  /**
+   * Checks if the user is not NONE or PENDING, and therefore on the team
+   *
+   * @param userId the id of the user
+   * @param teamId the id of the team
+   * @return a boolean indicating if they're on the team
+   */
+  private boolean isOnTeam(int userId, int teamId) {
```
liammoyn, line 41: Generally for these helper methods that are checking if a condition is met we can either return a boolean or throw and exception if a case isn't met, but we shouldn't combine the two. In this case, isOnTeam is something that could be reasonably used in a way where we wouldn't want to throw an exception if they're not on the team, so we should just return false instead.
Then this whole method can be reduced to
```
return db.fetchExists(
            db.selectFrom(USERS_TEAMS)
                    .where(USERS_TEAMS.USER_ID.eq(userId))
                    .and(USERS_TEAMS.TEAM_ID.eq(teamId))
                    .and(USERS_TEAMS.TEAM_ROLE.notEqual(TeamRole.None))
                    .and(USERS_TEAMS.TEAM_ROLE.notEqual(TeamRole.PENDING)));
```

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Checks if the block is open, meaning users can reserve it. Blocks are open if there is no
+   * record or if they've been released or uncompleted as the last action.
+   *
+   * @param blockId the id of the block to check
+   */
+  private void blockOpenCheck(int blockId) {
+    Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
+
+    if (maybeReservation.isPresent()) {
```
liammoyn, line 91: You can combine these if statements as well

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Checks if the block has been completed as its last action.
+   *
+   * @param blockId the id of the block to check
+   */
+  private void blockCompleteCheck(int blockId) {
+    Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
+
+    if (maybeReservation.isPresent()) {
```
liammoyn, line 140: Combine if statements

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Checks if the block was marked for QA in its last action.
+   *
+   * @param blockId the id of the block to check
+   */
+  private void blockQACheck(int blockId) {
+    Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
+
+    if (maybeReservation.isPresent()) {
```
liammoyn, line 157: If statements

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -0,0 +1,292 @@
...
+
+    if (userId != null && !db.fetchExists(db.selectFrom(USERS).where(USERS.ID.eq(userId)))) {
+        throw new UserDoesNotExistException(userId);
+    }
+
+    if (teamId != null && !db.fetchExists(db.selectFrom(TEAMS).where(TEAMS.ID.eq(teamId)))) {
+        throw new ResourceDoesNotExistException(teamId, "team");
+    }
+
+    if (userId != null && teamId != null) {
```
liammoyn, line 66: If we change isOnTeam to return true or false then you can just add it to the end of this if clause and throw the exception inside the if

api/src/main/java/com/codeforcommunity/dto/reservation/CompleteReservationRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "complete_reservation_request";
+    List<String> fields = new ArrayList<>();
+
+    if (blockID == null) {
+      fields.add(fieldName + "blockID");
+    }
+    if (teamID == null) {
```
liammoyn, line 43: Wait isn't it a valid use case to have a null teamID?

api/src/main/java/com/codeforcommunity/dto/reservation/MakeReservationRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "make_reservation_request";
+    List<String> fields = new ArrayList<>();
+
+    if (blockID == null) {
+      fields.add(fieldName + "blockID");
+    }
+    if (teamID == null) {
```
liammoyn, line 43: Is teamID allowed to be null?

api/src/main/java/com/codeforcommunity/api/IReservationProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+   * Marks the given block as reserved for the given user/team by creating a new entry in the
+   * reservation table
+   */
+  void makeReservation(JWTData userData, MakeReservationRequest makeReservationRequest);
+
+  /**
+   * Marks the given block as complete for the given user/team by creating a new entry in the
+   * reservation table
+   */
+  void completeReservation(JWTData userData, CompleteReservationRequest completeReservationRequest);
```
liammoyn, line 18: We should probably combine CompleteResrvationRequest and MakeReservationRequest as well... Something like BlockIDOptionalTeamIdRequest???

api/src/main/java/com/codeforcommunity/exceptions/IncorrectBlockStatusException.java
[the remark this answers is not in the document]
liammoyn: If you make a stub ticket and remind me later I'll show you, it's not a big implementation and it's a useful technique to know

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
florisdobber: Not as simple, leaving as is

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
florisdobber: Same as above

api/src/main/java/com/codeforcommunity/dto/reservation/MakeReservationRequest.java
florisdobber: Fixed

api/src/main/java/com/codeforcommunity/dto/reservation/CompleteReservationRequest.java
florisdobber: Fixed

api/src/main/java/com/codeforcommunity/api/IReservationProcessor.java
florisdobber: Feel like this is clearer for now and might receive changes when teams is implemented, don't want to have to undo it.

service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
florisdobber: Done

liammoyn (approved):
Good job getting through this

PR #5, opened 2020-12-27
Finish user router

florisdobber:
Added the last missing route for the user router, change privilege level.

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
+++ b/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
@@ -1,5 +1,6 @@
 package com.codeforcommunity.dto;
 
+import com.codeforcommunity.enums.PrivilegeLevel;
 import com.codeforcommunity.exceptions.HandledException;
 import com.codeforcommunity.exceptions.MalformedParameterException;
 import java.util.List;
@@ -101,5 +102,20 @@ public abstract class ApiDto {
    */
   protected boolean passwordInvalid(String pass) {
     return pass == null || pass.trim().length() < 8;
+  }
+
+  /**
+   *  Checks to see if the a privilege level is in the enum.
+   *
+   * @param level the privilege level to check
+   * @return a boolean representing whether the given privilege level is invalid or not.
+   */
+  protected boolean privilegeLevelInvalid(String level) {
+    for (PrivilegeLevel val : PrivilegeLevel.values()) {
+      if (val.name().equals(level)) {
+        return false;
+      }
+    }
+    return true;
   }
 }
--- a/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
+++ b/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
@@ -112,7 +112,7 @@ public abstract class ApiDto {
    */
   protected boolean privilegeLevelInvalid(String level) {
     for (PrivilegeLevel val : PrivilegeLevel.values()) {
-      if (val.name().equals(level)) {
+      if (val.name().toLowerCase().equals(level.toLowerCase())) {
         return false;
       }
     }
--- a/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
+++ b/api/src/main/java/com/codeforcommunity/dto/ApiDto.java
@@ -105,7 +105,7 @@ public abstract class ApiDto {
   }
 
   /**
-   *  Checks to see if the a privilege level is in the enum.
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/codeforcommunity/dto/ApiDto.java
```diff
@@ -102,4 +103,19 @@ protected boolean isEmpty(String str) {
...
+
+  /**
+   * Checks to see if the a privilege level is in the enum.
+   *
+   * @param level the privilege level to check
+   * @return a boolean representing whether the given privilege level is invalid or not.
+   */
+  protected boolean privilegeLevelInvalid(String level) {
+    for (PrivilegeLevel val : PrivilegeLevel.values()) {
+      if (val.name().equals(level)) {
```
liammoyn, line 115: Maybe do a .toUpperCase() or .toLowerCase() here just so no one ends up with weird capitalization errors

service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -95,4 +95,41 @@ public void changeEmail(JWTData userData, ChangeEmailRequest changeEmailRequest)
...
+    UsersRecord user =
+        db.selectFrom(USERS)
+            .where(USERS.EMAIL.eq(changePrivilegeLevelRequest.getTargetUserEmail()))
+            .fetchOne();
+    if (user == null) {
+      throw new UserDoesNotExistException(changePrivilegeLevelRequest.getTargetUserEmail());
+    }
+
+    // check if user is admin
+    if (!(userData.getPrivilegeLevel() == PrivilegeLevel.ADMIN
```
liammoyn, line 111: This check should always come first in the method. Otherwise we're letting non-admin users check any email they want to see if it exists in our db

api/src/main/java/com/codeforcommunity/dto/ApiDto.java
liammoyn: Ah I just realized superAdmin is camel case.... We probably want to change that to not depend on capitalization

service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -95,4 +95,41 @@ public void changeEmail(JWTData userData, ChangeEmailRequest changeEmailRequest)
...
+    // normal admins can't create super admins
+    if (userData.getPrivilegeLevel() == PrivilegeLevel.ADMIN
+        && changePrivilegeLevelRequest.getNewLevel() == PrivilegeLevel.SUPER_ADMIN) {
+      throw new AuthException("User does not have the required privilege level");
+    }
+
+    // check password and if the privilege level is different
+    if (Passwords.isExpectedPassword(
+        changePrivilegeLevelRequest.getPassword(), user.getPasswordHash())) {
+      if (db.fetchExists(
```
liammoyn, line 125: We can check this with the UsersRecord instance directly, this saves us from doing another trip to the database

liammoyn (changes requested):

api/src/main/java/com/codeforcommunity/dto/ApiDto.java
florisdobber: On the front end a drop down is used so capitalization should not be an issue since there's preset options

florisdobber: Made the change anyway since it shouldn't really matter

service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -95,4 +95,40 @@ public void changeEmail(JWTData userData, ChangeEmailRequest changeEmailRequest)
...
+    }
+
+    // normal admins can't create super admins
+    if (userData.getPrivilegeLevel() == PrivilegeLevel.ADMIN
+        && changePrivilegeLevelRequest.getNewLevel() == PrivilegeLevel.SUPER_ADMIN) {
+      throw new AuthException("User does not have the required privilege level");
+    }
+
+    // check password and if the privilege level is different
+    if (Passwords.isExpectedPassword(
```
liammoyn, line 123: I think this should be checking the changePrivilegeLevelRequest.getPassword() against the calling user's password, not the target user's password that's being checked here

liammoyn (approved):

PR #6, opened 2020-12-28
Did initial setup for import router

florisdobber:


api/src/main/java/com/codeforcommunity/dto/imports/ImportBlocksRequest.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public void setBlocks(List<BlockImport> blocks) {
+    this.blocks = blocks;
+  }
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String newFieldPrefix = fieldPrefix + "importNeighborhoodRequest.";
+
```
florisdobber, line 29: Not sure what is happening here but I will trust it works.

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class ImportProcessorImpl implements IImportProcessor {
+  private final DSLContext db;
+
+  public ImportProcessorImpl(DSLContext db) {
+    this.db = db;
+  }
+
+  @Override
+  public void importBlocks(JWTData userData, ImportBlocksRequest importBlocksRequest) {
```
florisdobber, line 24: Should there be any checks that there is data in the neighborhoods table before being able to run this route? Since blocks are foreign keyed to neighborhoods storing them without an existing neighborhood would throw an error I believe.

api/src/main/java/com/codeforcommunity/dto/imports/ImportBlocksRequest.java
liammoyn: Essentially this is just calling validateFields for each BlockImport in the this.blocks List. If any of those objects have nulls where they shouldn't then they'll show up as importNeighborhoodRequest.blockImport.fieldName.

The weird return value is making use of Java streams which are very useful when trying to apply a function to every item in a list, actually pretty similar to things you might have done in racket in fundies 1

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
liammoyn: That's true. The reason I didn't originally is because it's kind of unavoidable that we're exposing ourselves to foreign key exceptions unless we check each ID before we insert it. Having a baseline check that just checks that there's _something_ there could be helpful though, what do you think?

liammoyn (approved):

api/src/main/java/com/codeforcommunity/dto/imports/ImportBlocksRequest.java
florisdobber: I see, thanks for explaining!

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
florisdobber: I think a check for having something there is good and easy to implement. Checking each id seems like too much and a whole different issue if that is not adding up versus whether someone forgot to call the importNeighborhoods route before.

PR #8, opened 2021-01-04
Implemented routes for map to get block and neighborhood geo json

liammoyn:
Note that there is a mismatch in how our geometry field is stored and how it's translated to JSON. Currently, the seeding scripts use single quotes (') in the geometry but the JsonObject requires double quotes (") to correctly parse the JSON. Right now I'm doing a String.replaceAll but this is too slow for something that we're going to be calling pretty constantly.

Ideally, we can just change our seeding scripts to use " instead of ' and then we wouldn't have any problems. If, for some reason, this doesn't work then it's probably worth trying to do something more fancy with our JSON parsing to accommodate.

Also, jOOQ queries for getting completion percentage were very tough to figure out and I'm sure we'll have to make similar queries for different functionalities. Check out the getNeighborhoodCompletionPercentage function or ask me if you want an explanation of how we should handle getting multiple blocks that meet a reservation status.


florisdobber (approved):
Adjust the seeding scripts seems easiest correct? You want to do that in this PR or should we create a separate PR for that?

As for the completion percentage, the logic is what I envisioned it to be like, just turned out to be more complicated than anticipated. We might have some similar queries but I think based on this logic other people should be able to figure out how to do it.

PR #9, opened 2021-01-04
Fixed signup to include username

liammoyn:
We want usernames so that we can show them on leaderboards, this adds them back to signup


florisdobber (approved):
Looks good, thanks for making the change!

PR #11, opened 2021-01-17
Added import reservations route

florisdobber:


persist/src/main/resources/db/seed/reservations.json
```diff
@@ -0,0 +1,5653 @@
+{
+  "reservations": [
+    {
+      "blockId": 439,
+      "userId": 47.0,
```
liammoyn, line 5: Weird that these ids have `.0` but as long as you tested that this doesn't mess anything up it's fine

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
```diff
@@ -61,4 +70,48 @@ public void importNeighborhoods(
...
+
+    for (ReservationImport reservationImport : importReservationsRequest.getReservations()) {
+      if (!db.fetchExists(
+          db.selectFrom(BLOCKS).where(BLOCKS.ID.eq(reservationImport.getBlockId())))) {
+        throw new ResourceDoesNotExistException(reservationImport.getBlockId(), "block");
+      }
+      if (reservationImport.getUserId() != null
+          && (!db.fetchExists(
+              db.selectFrom(USERS).where(USERS.ID.eq(reservationImport.getUserId()))))) {
+        throw new UserDoesNotExistException(reservationImport.getUserId());
```
liammoyn, line 92: It's weird that we have a special case for users instead of just using the ResourceDoesNotExistException, we might want to simplify and just use the same one for everything. This can be done in a followup PR though

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
```diff
@@ -61,4 +70,48 @@ public void importNeighborhoods(
...
+  public void importReservations(
+      JWTData userData, ImportReservationsRequest importReservationsRequest) {
+    if (userData.getPrivilegeLevel() != PrivilegeLevel.SUPER_ADMIN) {
+      throw new AuthException("User does not have the required privilege level.");
+    }
+
+    // Get superuser ID
+    Integer superAdminId = userData.getUserId();
+
+    for (ReservationImport reservationImport : importReservationsRequest.getReservations()) {
```
liammoyn, line 84: So right now if there is some invalid entry halfway through the list (because a block id doesn't exist or something) then the first half of the list will be in the DB and the second half will be completely ignored. This would be a nightmare to try to rectify since then you would have to manually cut out all the reservations that were successful and remake the request body.
Instead what we can do is first translate each ReservationImport into a ReservationsRecord while checking their validity and then separately storing them once we know they're all valid. This way it's all or nothing and it's much easier to fix and rerun. This would look something like this:
```java
List<ReservationsRecord> records = importReservationsRequest.getReservations().map(reservationImport -> {
  if (...) { throw new ResourceDoesNotExistException(...) }
  ...
  ReservationsRecord reservation = db.newRecord(RESERVATIONS);
  ...
  return reservation;
});
records.forEach(record -> {
  record.store();
});
```

persist/src/main/resources/db/seed/reservations.json
florisdobber: Yeah idk why either but it doesn't break anything

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
florisdobber: Considered it but the reason I left it in for now is that the user exception can take both an ID or username which I liked

service/src/main/java/com/codeforcommunity/processor/ImportProcessorImpl.java
florisdobber: Fixed

liammoyn (approved):

PR #12, opened 2021-01-17
New teams router

florisdobber:
Created the teams router but accidentally wrote the wrong name in the commits/branch


api/src/main/java/com/codeforcommunity/exceptions/TeamLeaderOnlyRouteException.java
```diff
@@ -0,0 +1,22 @@
+package com.codeforcommunity.exceptions;
+
+import com.codeforcommunity.rest.FailureHandler;
+import io.vertx.ext.web.RoutingContext;
+
+public class TeamLeaderOnlyRouteException extends HandledException {
```
liammoyn, line 6: Maybe call this wrong TeamRoleException and take a TeamRole value(s) so that we can reuse it down the line

persist/src/main/resources/db/migration/V1__Initial_Import.sql
```diff
@@ -39,7 +39,8 @@ CREATE TABLE IF NOT EXISTS teams
     team_name   VARCHAR(100)    NOT NULL,
     bio         TEXT            NOT NULL,
     finished    BOOLEAN         NOT NULL DEFAULT FALSE,
-    created_at  TIMESTAMP       NOT NULL DEFAULT CURRENT_TIMESTAMP
+    created_at  TIMESTAMP       NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    deleted_at  TIMESTAMP
```
liammoyn, line 43: Fine for now but be careful changing these files

florisdobber: Yeah, this was an exception since it should've been done before really but I'd missed it. All the tree/site db stuff will go in a second file

api/src/main/java/com/codeforcommunity/exceptions/TeamLeaderOnlyRouteException.java
florisdobber: Done

liammoyn (approved):

PR #14, opened 2021-01-28
Add unit tests for reservation processor

liammoyn:
To test the `basicChecks()` method the JooqMock class has to be updated to support mocking fetchExists calls.

I couldn't get maven to properly run the `setup()` method with the `@BeforeEach` annotation. I mention in a TODO but the solution is in [this stack overflow post](https://stackoverflow.com/questions/51382356/maven-does-not-run-beforeeach-methods-while-running), the problem is we currently don't specify our maven surefire plugin explicitly and I am too lazy to tackle that right now.

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
@@ -26,7 +26,7 @@ public class ReservationProcessorImpl implements IReservationProcessor {
     this.db = db;
   }
 
-  public Optional<ReservationsRecord> lastAction(int blockId) {
+  Optional<ReservationsRecord> lastAction(int blockId) {
     return Optional.ofNullable(
         db.selectFrom(RESERVATIONS)
             .where(RESERVATIONS.BLOCK_ID.eq(blockId))
@@ -42,7 +42,7 @@ public class ReservationProcessorImpl implements IReservationProcessor {
    * @param teamId the id of the team
    * @return a boolean indicating if they're on the team
    */
-  private boolean isOnTeam(int userId, int teamId) {
+  boolean isOnTeam(int userId, int teamId) {
     return db.fetchExists(
         db.selectFrom(USERS_TEAMS)
             .where(USERS_TEAMS.USER_ID.eq(userId))
@@ -51,7 +51,8 @@ public class ReservationProcessorImpl implements IReservationProcessor {
             .and(USERS_TEAMS.TEAM_ROLE.notEqual(TeamRole.PENDING)));
   }
 
-  private void basicChecks(int blockId, Integer userId, Integer teamId) {
+  // TODO: This method needs to be tested
+  void basicChecks(int blockId, Integer userId, Integer teamId) {
     if (!db.fetchExists(db.selectFrom(BLOCKS).where(BLOCKS.ID.eq(blockId)))) {
       throw new ResourceDoesNotExistException(blockId, "block");
     }
@@ -76,7 +77,7 @@ public class ReservationProcessorImpl implements IReservationProcessor {
    *
    * @param level the privilege level of the user calling the route
    */
-  private void isAdminCheck(PrivilegeLevel level) {
+  void isAdminCheck(PrivilegeLevel level) {
     if (!(level.equals(PrivilegeLevel.ADMIN) || level.equals(PrivilegeLevel.SUPER_ADMIN))) {
       throw new AuthException("User does not have the required privilege level.");
     }
@@ -88,7 +89,7 @@ public class ReservationProcessorImpl implements IReservationProcessor {
    *
    * @param blockId the id of the block to check
    */
-  private void blockOpenCheck(int blockId) {
+  void blockOpenCheck(int blockId) {
     Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
 
     if (maybeReservation.isPresent()
@@ -104,7 +105,7 @@ public class ReservationProcessorImpl implements IReservationProcessor {
    * @param blockId the id the block to check
    * @param userId the id of the user calling the route
    */
-  private void blockReservedCheck(int blockId, int userId) {
+  void blockReservedCheck(int blockId, int userId) {
     Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
 
     // check if there are any entries
@@ -138,14 +139,11 @@ public class ReservationProcessorImpl implements IReservationProcessor {
    *
    * @param blockId the id of the block to check
    */
-  private void blockCompleteCheck(int blockId) {
+  void blockCompleteCheck(int blockId) {
     Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
 
-    if (maybeReservation.isPresent()) {
-      if (!(maybeReservation.get().getActionType().equals(ReservationAction.COMPLETE))) {
-        throw new IncorrectBlockStatusException(blockId, "complete");
-      }
-    } else {
+    if (!maybeReservation.isPresent()
+        || !(maybeReservation.get().getActionType().equals(ReservationAction.COMPLETE))) {
       throw new IncorrectBlockStatusException(blockId, "complete");
     }
   }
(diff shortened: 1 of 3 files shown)


service/src/main/java/com/codeforcommunity/processor/ReservationProcessorImpl.java
```diff
@@ -138,14 +139,11 @@ private void blockReservedCheck(int blockId, int userId) {
    *
    * @param blockId the id of the block to check
    */
-  private void blockCompleteCheck(int blockId) {
+  void blockCompleteCheck(int blockId) {
     Optional<ReservationsRecord> maybeReservation = lastAction(blockId);
 
-    if (maybeReservation.isPresent()) {
-      if (!(maybeReservation.get().getActionType().equals(ReservationAction.COMPLETE))) {
-        throw new IncorrectBlockStatusException(blockId, "complete");
```
connernilsen, line 146: lmao

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+  ReservationProcessorImpl proc;
+
+  /**
+   * Method to setup mockDb and processor
+   *
+   * <p>TODO: This can't be run with the @BeforeEach annotation by maven currently. The way to
+   * enable this is outlined in this Stack Overflow post:
+   * https://stackoverflow.com/questions/51382356/maven-does-not-run-beforeeach-methods-while-running
+   * This isn't currently being done because our version of maven surefire is auto imported and is
+   * only version 2.12.4
```
connernilsen, line 33: Hmm, it used to be explicitly imported a while ago. I wonder why it got removed

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+      this.proc.isAdminCheck(PrivilegeLevel.SUPER_ADMIN);
+    } catch (AuthException e) {
+      fail("Method should not have thrown an exception");
+    }
+  }
+
+  @Test
+  public void testBlockOpenCheckNoAction() {
+    setup();
+    doReturn(Optional.empty()).when(this.proc).lastAction(anyInt());
```
connernilsen, line 74: This is more of a nitpickey thing, but I prefer the `when(mock.doSomething()).thenReturn(somethingElse)` style since it enforces strict typing. I guess it probably doesn't matter as much in tests though.

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  @Test
+  public void testBlockOpenCheckUncomplete() {
+    setup();
+    ReservationsRecord mockReservation = mock(ReservationsRecord.class);
+    doReturn(Optional.of(mockReservation)).when(this.proc).lastAction(anyInt());
+
+    when(mockReservation.getActionType())
+        .thenReturn(ReservationAction.UNCOMPLETE, ReservationAction.UNCOMPLETE);
```
connernilsen, line 90: `UNCOMPLETE`? Not `INCOMPLETE`? I know it's not part of your PR, but I just wanted to see what's up with it

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+      fail();
+    } catch (IncorrectBlockStatusException e) {
+      assertEquals(1, e.getBlockId());
+      assertEquals("reserved", e.getExpectedStatus());
+    }
+
+    verify(mockReservation, times(1)).getActionType();
+    verify(mockReservation, times(0)).getUserId();
+    verify(mockReservation, times(0)).getTeamId();
+    verify(this.proc, times(0)).isOnTeam(anyInt(), anyInt());
```
connernilsen, line 153: Instead of doing `times(0)`, it might be better to do `never()`

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+
+    try {
+      this.proc.blockReservedCheck(1, 15);
+      fail();
+    } catch (IncorrectBlockStatusException e) {
+      assertEquals(1, e.getBlockId());
+      assertEquals("reserved", e.getExpectedStatus());
+    }
+
+    verify(mockReservation, times(1)).getActionType();
```
connernilsen, line 150: `times(1)` is default

connernilsen (commented):
It's unrelated to your PR, but the README still talks about being the backend-scaffold. Just wanted to point it out since I'll forget soon

florisdobber (commented):
Thanks for setting this up. It's super informative to see how it works and we can then build from here! I'll leave Conner to approve this since I don't actually know what it's supposed to look like.

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
liammoyn: You can't do that with spies (which this.proc is) since it calls the actual method.

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
liammoyn: I think it was specified as uncomplete because it comes up when a completed block is reset to incomplete... so uncomplete is a verb. Probably still not grammatically correct though

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
liammoyn: Probably but I liked the consistent format

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
liammoyn: Again, liked the consistent format

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
connernilsen: Wait, why are we spying on the thing we're testing?

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
connernilsen: Yeah it just reads a little weird to me. Probably not worth it to go through the effort to change it though.

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
```diff
@@ -0,0 +1,330 @@
...
+    } catch (IncorrectBlockStatusException e) {
+      fail("Method should not throw exception");
+    }
+  }
+
+  @Test
+  public void testBlockOpenCheckReserved() {
+    setup();
+    ReservationsRecord mockReservation = mock(ReservationsRecord.class);
+    doReturn(Optional.of(mockReservation)).when(this.proc).lastAction(anyInt());
```
connernilsen, line 103: Similarly to above, in the case where you need to return a valid value, you can call `mockDb.addReturn("SELECT", mockReservation)` (or `mockDb.addReturn(OperationType.SELECT, mockReservation)` after the new `JooqMock` PR gets merged)

connernilsen (changes requested):
Whoops, I missed a few important things the first round that seem off to me

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
connernilsen: Sure

service/src/test/java/com.codeforcommunity/processor/ReservationProcessorImplTest.java
liammoyn: Spying a tested class is valid as long as you're calling the real method you're testing and you test the methods you spy separately. I would argue it's better practice even since it makes your unit tests more contained so that they only test precisely what's in the method you're testing

PR #15, opened 2021-01-29
Changed leaderboard routes to have query params instead of request bo…

liammoyn:
…dies


florisdobber (approved):
Looks good, thanks for making the change!

PR #16, opened 2021-02-01
Add Teams Routes

oloveluck:
Draft PR. More to come soon


api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
```diff
@@ -1,7 +1,16 @@
 package com.codeforcommunity.api;
 
 import com.codeforcommunity.auth.JWTData;
+import com.codeforcommunity.dto.team.*;
```
florisdobber, line 4: Can you make sure everything is explicitly imported and not with a *? You can make sure IntelliJ doesn't do this automatically in the settings somewhere

api/src/main/java/com/codeforcommunity/exceptions/TeamDoesNotExistException.java
```diff
@@ -0,0 +1,21 @@
+package com.codeforcommunity.exceptions;
```
florisdobber, line 1: Can you use resourceDoesNotExist exception for this?

api/src/main/java/com/codeforcommunity/rest/FailureHandler.java
```diff
@@ -176,6 +176,11 @@ private void handleUncaughtError(RoutingContext ctx, Throwable throwable) {
     end(ctx, message, 500);
   }
 
+  public void handleInvalidTeamException(RoutingContext ctx, TeamDoesNotExistException e) {
```
florisdobber, line 179: So this wouldn't be necessary

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -1,14 +1,16 @@
 package com.codeforcommunity.processor;
 
-import static org.jooq.generated.Tables.TEAMS;
-import static org.jooq.generated.Tables.USERS_TEAMS;
+import static org.jooq.generated.Tables.*;
```
florisdobber, line 3: Same thing here

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -1,14 +1,16 @@
...
 
 import com.codeforcommunity.api.ITeamsProcessor;
 import com.codeforcommunity.auth.JWTData;
+import com.codeforcommunity.dto.team.*;
```
florisdobber, line 7: And here

api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
liammoyn: Preferences > Editor > Code Style > Java > Imports: Change number of lines for * to 999

api/src/main/java/com/codeforcommunity/dto/team/AddGoalRequest.java
```diff
@@ -0,0 +1,74 @@
...
+    this.goal = goal;
+    this.teamId = teamId;
+    this.startAt = startAt;
+    this.completeBy = completeBy;
+  }
+
+  private AddGoalRequest() {}
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
```
liammoyn, line 25: This is super nitpicky but let's be consistent about where in the DTO class this method goes. I think by convention we've been doing it as the last thing in the class

api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
```diff
@@ -1,7 +1,16 @@
...
 public interface ITeamsProcessor {
   void disbandTeam(JWTData userData, int teamId);
+
+  void createTeam(JWTData userData, CreateTeamRequest request);
+
+  TeamDataResponse getTeam(JWTData userData, GetTeamRequest getTeamRequest);
+
+  void addGoal(JWTData userData, AddGoalRequest addGoalRequest);
+
+  void deleteGoal(JWTData userData, DeleteGoalRequest deleteGoalRequest);
```
liammoyn, line 15: Delete goal doesn't have any JSON body and just has the goalId in the path. For these cases, creating a DTO doesn't save us any time because we don't use that `jsonBodyToDTO` method in `RestFunctions`. So instead of creating a new DTO just add a `Integer goalId` parameter to this interface instead

api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
```diff
@@ -1,7 +1,16 @@
...
 
 import com.codeforcommunity.auth.JWTData;
+import com.codeforcommunity.dto.team.*;
 
 public interface ITeamsProcessor {
   void disbandTeam(JWTData userData, int teamId);
+
+  void createTeam(JWTData userData, CreateTeamRequest request);
+
+  TeamDataResponse getTeam(JWTData userData, GetTeamRequest getTeamRequest);
```
liammoyn, line 11: Same with deleteGoal, no JSON body for getTeam so just use a `Integer teamId` parameter instead of creating a DTO for it

api/src/main/java/com/codeforcommunity/dto/team/DeleteGoalRequest.java
```diff
@@ -0,0 +1,12 @@
+package com.codeforcommunity.dto.team;
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.List;
+
+public class DeleteGoalRequest extends ApiDto {
```
liammoyn, line 7: After changing the Interface method signatures this class can be deleted

api/src/main/java/com/codeforcommunity/dto/team/GetTeamRequest.java
```diff
@@ -0,0 +1,35 @@
+package com.codeforcommunity.dto.team;
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class GetTeamRequest extends ApiDto {
```
liammoyn, line 8: After changing the Interface method signatures this class can be deleted

api/src/main/java/com/codeforcommunity/dto/team/TeamDataResponse.java
```diff
@@ -0,0 +1,47 @@
+package com.codeforcommunity.dto.team;
+
+import java.sql.Timestamp;
+
+public class TeamDataResponse {
+  private final Integer id;
```
liammoyn, line 6: I think you forgot to create a getter for this field

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +26,71 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+    registerCreateTeam(router);
+    registerGetTeam(router);
+    registerAddGoal(router);
+    registerDeleteGoal(router);
     registerDisbandTeam(router);
 
     return router;
   }
 
+  // Create a Team
```
liammoyn, line 38: Whenever possible try using java doc for comments above classes, methods, or fields. Java doc is just a comment in the form of `/** (whatever you want to say) */`. This doesn't really matter too much for us since we're not going to be generating documentation from our code but it is a good habit to get into

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +26,71 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+  // Get a Team
+  private void registerGetTeam(Router router) {
+    Route getTeamRoute = router.get("/:team_id");
+    getTeamRoute.handler(this::handleGetTeamRoute);
+  }
+
+  private void handleGetTeamRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    GetTeamRequest getTeamRequest =
+        RestFunctions.getJsonBodyAsClass(routingContext, GetTeamRequest.class);
```
liammoyn, line 61: Because the teamId is in the URL instead of in a JSON body in the spec use the getRequestParameterAsInt method in RestFunctions. The name for the route is the same thing as in line 54, so you would pass "team_id"

api/src/main/java/com/codeforcommunity/dto/team/AddGoalRequest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.security.Timestamp;
+import java.util.ArrayList;
+import java.util.List;
+
+public class AddGoalRequest extends ApiDto {
+  private Integer goal;
+  private Integer teamId;
```
liammoyn, line 11: Generally, request DTOs should only contain the info that is in the JSON body. In this case, the API specification has teamId in the path as a path parameter. To support that, you should take this out of the DTO and add another argument to the addGoal interface method.

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +26,71 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+  private void registerAddGoal(Router router) {
+    Route addGoalRoute = router.post("/:team_id/add_goal");
+    addGoalRoute.handler(this::handleAddGoal);
+  }
+
+  private void handleAddGoal(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jtw_data");
+    AddGoalRequest addGoalRequest =
+        RestFunctions.getJsonBodyAsClass(routingContext, AddGoalRequest.class);
+    processor.addGoal(userData, addGoalRequest);
```
liammoyn, line 76: Once you take the teamId out of the DTO then you'll have to have another line getting the team_id from the path (it's the same as in the getTeam route) and then passing it to the processor along with the DTO.

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +37,46 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+        team.getDeletedAt());
+  }
+
+  @Override
+  public void addGoal(JWTData userData, AddGoalRequest addGoalRequest) {
+    // TODO if user is not leader. thror error and send 401 unauthorized
+    GoalsRecord goal = db.newRecord(GOALS);
+    goal.setGoal(addGoalRequest.getGoal());
+    goal.getStartAt(addGoalRequest.getCompleteBy());
+    goal.getCompleteBy(addGoalRequest.getCompleteBy());
```
liammoyn, line 77: Don't forget to call `goal.store()` otherwise jOOQ is never going to add this record to the database

service/src/main/java/com/codeforcommunity/requester/Emailer.java
```diff
@@ -93,4 +93,16 @@ public void sendAccountDeactivatedEmail(String sendToEmail, String sendToName) {
     emailBody.ifPresent(
         s -> emailOperations.sendEmail(sendToName, sendToEmail, subjectAccountDeleted, s));
   }
+
+  public void sendInviteTeamEmail(String sendToEmail, String sendToName, String teamName) {
```
liammoyn, line 97: Add a `// TODO` comment above this method so we don't forget to come back to it.

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+    end(routingContext.response(), 200);
+  }
+
+  // Get a Team
+  private void registerGetTeam(Router router) {
+    Route getTeamRoute = router.get("/:team_id");
+    getTeamRoute.handler(this::handleGetTeamRoute);
+  }
+
+  private void handleGetTeamRoute(RoutingContext routingContext) {
```
florisdobber, line 68: Our convention is to call the RoutingContext ctx. Find and replace routingContext with ctx would fix that easily

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -36,9 +167,46 @@ private void registerDisbandTeam(Router router) {
...
+
+  private void handleLeaveTeamRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "team_id");
+    processor.leaveTeam(userData, teamId);
+    end(routingContext.response(), 200);
+  }
+
+  private void registerTransferTeamOwnership(Router router) {
+    Route transferOwnershipRoute = router.post("/:team_id");
```
florisdobber, line 200: Route should be "/:team_id/transfer_ownership"

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+  private void registerInviteUser(Router router) {
+    Route inviteUserRoute = router.post("/:team_id/invite");
+    inviteUserRoute.handler(this::handleInviteUser);
+  }
+
+  private void handleInviteUser(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    InviteUserRequest inviteUserRequest =
+        RestFunctions.getJsonBodyAsClass(routingContext, InviteUserRequest.class);
+    processor.inviteUser(userData, inviteUserRequest);
```
florisdobber, line 113: No response

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+
+  private void registerGetApplicants(Router router) {
+    Route getApplicantsRoute = router.get("/:team_id/applicants");
+    getApplicantsRoute.handler(this::handleGetApplicantsRoute);
+  }
+
+  private void handleGetApplicantsRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "team_id");
+    processor.getApplicants(userData, teamId);
```
florisdobber, line 124: Missing response

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+
+  private void registerApplyToTeam(Router router) {
+    Route applyToTeamRoute = router.post("/:team_id/apply");
+    applyToTeamRoute.handler(this::handleApplyToTeamRoute);
+  }
+
+  private void handleApplyToTeamRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "team_id");
+    processor.applyToTeam(userData, teamId);
```
florisdobber, line 135: Missing response

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+  private void registerApproveUser(Router router) {
+    Route approveUserRoute = router.post("/:team_id/applicants/:user_id/approve");
+    approveUserRoute.handler(this::handleApproveUserRoute);
+  }
+
+  private void handleApproveUserRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "team_id");
+    int memberId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "user_id");
+    processor.approveUser(userData, teamId, memberId);
```
florisdobber, line 147: Missing response

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,137 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+  private void registerRejectUserRoute(Router router) {
+    Route rejectUserRoute = router.post("/:team_id/applicants/:user_id/reject");
+    rejectUserRoute.handler(this::handleRejectUserRoute);
+  }
+
+  private void handleRejectUserRoute(RoutingContext routingContext) {
+    JWTData userData = routingContext.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "team_id");
+    int memberId = RestFunctions.getRequestParameterAsInt(routingContext.request(), "user_id");
+    processor.rejectUser(userData, teamId, memberId);
```
florisdobber, line 159: Missing response

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+  }
+
+  @Override
+  public TeamDataResponse getTeam(JWTData userData, Integer teamId) {
+    TeamsRecord team = db.selectFrom(TEAMS).where(TEAMS.ID.eq(teamId)).fetchOne();
+    if (team == null) {
+      throw new ResourceDoesNotExistException(teamId, "Team");
+    }
+
+    return new TeamDataResponse(
```
florisdobber, line 63: Missing goals of the team

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+        team.getTeamName(),
+        team.getBio(),
+        team.getFinished(),
+        team.getCreatedAt(),
+        team.getDeletedAt());
+  }
+
+  @Override
+  public void addGoal(JWTData userData, AddGoalRequest addGoalRequest, int teamId) {
+    UsersTeamsRecord usersTeamsRecord =
```
florisdobber, line 74: Check if team exists?

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+      goal.setStartAt(addGoalRequest.getStart_at());
+      goal.setCompleteBy(addGoalRequest.getComplete_by());
+      goal.store();
+    } else {
+      throw new WrongTeamRoleException(teamId, usersTeamsRecord.getTeamRole());
+    }
+  }
+
+  @Override
+  public void deleteGoal(JWTData userData, int teamId, int goalId) {
```
florisdobber, line 93: Check if team and goal exists?

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+
+    if (usersTeamsRecord != null && usersTeamsRecord.getTeamRole() == TeamRole.LEADER) {
+      db.deleteFrom(GOALS).where(GOALS.ID.eq(goalId));
+    } else {
+      throw new WrongTeamRoleException(teamId, usersTeamsRecord.getTeamRole());
+    }
+  }
+
+  @Override
+  public void inviteUser(JWTData userData, InviteUserRequest inviteUserRequest) {
```
florisdobber, line 108: Liam should take a look at this

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+    } else {
+      throw new WrongTeamRoleException(
+          inviteUserRequest.getTeamId(), inviterUserTeamsRecord.getTeamRole());
+    }
+  }
+
+  @Override
+  public List<UsersTeamDataResponse> getApplicants(
+      JWTData userData, int teamId) {
+    List<UsersTeamsRecord> applicants =
```
florisdobber, line 152: Check if team exists

florisdobber: Can probably extract this out to a simple function that takes a team ID and run it at the start of a lot of these functions

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +39,266 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+          UsersTeamDataResponse response =
+              new UsersTeamDataResponse(app.getUserId(), teamId, app.getTeamRole());
+          responses.add(response);
+        });
+    return responses;
+  }
+
+  @Override
+  public void applyToTeam(JWTData userData, int teamId) {
+    if (db.fetchExists(
```
florisdobber, line 166: Team exists check

florisdobber (commented):
Lots of moving code around and small forgotten things, the rest looks great! Looked over every file completely with the exception of TeamsProcessorImpl, will get to that another time

Can you remove the TeamDoesNotExistException by the way?

api/src/main/java/com/codeforcommunity/dto/team/UsersTeamDataResponse.java
```diff
@@ -0,0 +1,39 @@
+package com.codeforcommunity.dto.team;
```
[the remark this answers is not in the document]
oloveluck: During GetApplicants

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
oloveluck: oops forgot to push a final commit here...

api/src/main/java/com/codeforcommunity/dto/team/UsersTeamDataResponse.java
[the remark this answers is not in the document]
florisdobber: Got it thanks!

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+  @Override
+  public List<UsersTeamDataResponse> getApplicants(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
+    List<UsersTeamsRecord> applicants =
+        db.selectFrom(USERS_TEAMS).where(USERS_TEAMS.TEAM_ROLE.eq(TeamRole.PENDING)).fetch();
+    List<UsersTeamDataResponse> responses = new ArrayList<>();
+    applicants.forEach(
+        app -> {
+          UsersTeamDataResponse response =
+              new UsersTeamDataResponse(app.getUserId(), teamId, app.getTeamRole());
```
liammoyn, line 198: Based on the API Specification this DTO should just take the `app.getUserId()` and `app.getUsername()`. It doesn't make sense to include teamId and teamRole because they're the same for every applicant

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+    if (db.fetchExists(
+        USERS_TEAMS.where(
+            USERS_TEAMS
+                .TEAM_ID
+                .eq(teamId)
+                .and(USERS_TEAMS.USER_ID.eq(userData.getUserId()))))) {
+      throw new MemberApplicationException(
+          teamId, userData.getUserId());
+    } else {
+      UsersRecord invitedUser =
```
liammoyn, line 216: We can change this to a boolean gotten from a fetchExist. Then we can call it applyingUserOnTeam

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+  @Override
+  public void approveUser(JWTData userData, int teamId, int memberId) {
+    checkUserExists(memberId);
+    checkTeamExists(teamId);
+    UsersTeamsRecord leaderTeamsRecord =
+        db.selectFrom(USERS_TEAMS)
+            .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+            .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+            .fetchOne();
+    if (leaderTeamsRecord != null && leaderTeamsRecord.getTeamRole() == TeamRole.LEADER) {
```
liammoyn, line 239: Move role check into db call for the next 3 methods

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+      if (approvedUserRecord.getTeamRole() == TeamRole.PENDING) {
+        approvedUserRecord.setTeamRole(TeamRole.MEMBER);
+        approvedUserRecord.update();
+      } else {
+        throw new MemberStatusException(
+            memberId, teamId);
+      }
+    } else {
+      throw new WrongTeamRoleException(
+          leaderTeamsRecord.getTeamId(), leaderTeamsRecord.getTeamRole());
```
liammoyn, line 252: WrongTeamRoleException takes the role that the user is **supposed to** have not what they do have. Anywhere where you call WrongTeamRoleException make sure you're passing in a static role instead of one from the user

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+
+  @Override
+  public void leaveTeam(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
+    UsersTeamsRecord usersTeamsRecord =
+        db.selectFrom(USERS_TEAMS)
+            .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+            .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+            .fetchOne();
+    if (usersTeamsRecord != null && usersTeamsRecord.getTeamRole() != TeamRole.LEADER) {
```
liammoyn, line 315: This is probably a case where we would want to have a message if a leader tries to leave their own team that's different from the message you get if you're not on the team at all. You can create an improvement ticket for this

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+            .fetchOne();
+    UsersTeamsRecord newLeaderTeamsRecord =
+        db.selectFrom(USERS_TEAMS)
+            .where(
+                USERS_TEAMS
+                    .USER_ID
+                    .eq(transferOwnershipRequest.getNewLeaderId())
+                    .and(USERS_TEAMS.TEAM_ID.eq(teamId)))
+            .fetchOne();
+    if (oldLeaderTeamsRecord != null
```
liammoyn, line 340: Move both role checks into db calls

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+        && newLeaderTeamsRecord != null
+        && oldLeaderTeamsRecord.getTeamRole() == TeamRole.LEADER
+        && oldLeaderTeamsRecord.getTeamRole() == TeamRole.MEMBER) {
```
liammoyn, line 343: typo, should be newLeaderTeamsRecord

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+      newLeaderTeamsRecord.setTeamRole(TeamRole.LEADER);
+      oldLeaderTeamsRecord.setTeamRole(TeamRole.MEMBER);
+      newLeaderTeamsRecord.update();
+      oldLeaderTeamsRecord.update();
+    } else {
+      throw new WrongTeamRoleException(teamId, TeamRole.LEADER);
```
liammoyn, line 349: We need two different exceptions, one if the calling user isn't a team leader and one if the newLeader isn't a team member

liammoyn (changes requested):
Great progress on this! You got through the hard part so now it's just iteration

Also, you added an empty file that I can't comment on but make sure to delete that: `service/src/main/java/com/codeforcommunity/processor/TeamDoesNotExistException.java`

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
```diff
@@ -0,0 +1,58 @@
...
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class InviteUserRequest extends ApiDto {
+
+  private Integer teamId;
+  private Map<String, String> users;
```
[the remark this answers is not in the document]
oloveluck: Do we really want to validate the fields of each of the InviteContacts? This would result in an error being thrown if any of the users have an invalid email or leave a value none. The API Spec says in this case an email would just not be sent, and nothing will be changed in the database.

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
     team.setDeletedAt(new Timestamp(System.currentTimeMillis()));
     team.store();
   }
+
+  @Override
+  public void createTeam(JWTData userData, CreateTeamRequest request) {
+    TeamsRecord team = db.newRecord(TEAMS);
+    team.setBio(request.getBio());
+    team.setTeamName(request.getName());
```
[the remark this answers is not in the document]
oloveluck: When making tickets, is it best to make these isssues on github or put it in the monday backlog?

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,279 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+    checkGoalExists(goalId);
+
+    UsersTeamsRecord usersTeamsRecord =
+        db.selectFrom(USERS_TEAMS)
+            .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+            .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+            .fetchOne();
+
+    if (usersTeamsRecord != null && usersTeamsRecord.getTeamRole() == TeamRole.LEADER) {
+      db.deleteFrom(GOALS).where(GOALS.ID.eq(goalId));
```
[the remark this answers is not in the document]
oloveluck: I think deleting goals is better. Don't see why someone would want to restore a goal rather than just recreate it. We already have the ability to mark as complete, don't see why completed and deleted states are both necessary.

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
oloveluck: I created a different Exception for this

api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
```diff
@@ -1,7 +1,39 @@
...
+
+  void approveUser(JWTData userData, int teamId, int memberId);
+
+  void rejectUser(JWTData userData, int teamId, int memberId);
+
+  void kickUser(JWTData userData, int teamId, int memberId);
+
+  void leaveTeam(JWTData userData, int teamId);
+
+  void transferOwnership(JWTData userData, TransferOwnershipRequest transferOwnershipRequest, int teamId);
```
jackblanc, line 38: woah, we should absolutely do each route one at a time going forwards, this is over 1000 lines of code which is very hard to review

api/src/main/java/com/codeforcommunity/dto/team/GoalResponse.java
```diff
@@ -0,0 +1,39 @@
...
+import java.sql.Timestamp;
+
+public class GoalResponse {
+    private final int id;
+    private final int goal;
+    private final Timestamp startDate;
+    private final Timestamp completeBy;
+    private final Timestamp completionDate;
+
+    public GoalResponse(int id, int goal, Timestamp startDate, Timestamp completeBy, Timestamp completionDate) {
```
jackblanc, line 12: Is this ID for the goal or the team?

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
```diff
@@ -0,0 +1,38 @@
+package com.codeforcommunity.dto.team;
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class InviteUserRequest extends ApiDto {
```
jackblanc, line 9: ⛏️ InviteUsersRequest, for the list<users>

api/src/main/java/com/codeforcommunity/dto/team/TeamDataResponse.java
```diff
@@ -0,0 +1,67 @@
+package com.codeforcommunity.dto.team;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+
+import java.sql.Timestamp;
+import java.util.List;
+
+public class TeamDataResponse {
+  private final Integer id;
+  private final String teamName;
```
jackblanc, line 10: There's usages of lowerCamelCase and snake_case throughout different files. Which is more prevalent in this codebase? I personally prefer lowerCamelCase as that's what we use on the frontend.

api/src/main/java/com/codeforcommunity/dto/team/TransferOwnershipRequest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransferOwnershipRequest extends ApiDto {
+  private Integer newLeaderId;
+
+  public TransferOwnershipRequest(Integer newLeaderId) {
```
jackblanc, line 11: ❓ `newLeaderUserId` might be more clear

api/src/main/java/com/codeforcommunity/exceptions/LeaderCannotLeaveTeamException.java
```diff
@@ -0,0 +1,27 @@
+package com.codeforcommunity.exceptions;
+
+import com.codeforcommunity.rest.FailureHandler;
+import io.vertx.ext.web.RoutingContext;
+
+public class LeaderCannotLeaveTeamException extends HandledException {
```
jackblanc, line 6: To clarify this, as I commented on the dev docs PR, can a leader leave if there is more than one leader?

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,143 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+    createTeamRoute.handler(this::handleCreateTeamRoute);
+  }
+
+  private void handleCreateTeamRoute(RoutingContext ctx) {
+    JWTData userData = ctx.get("jwt_data");
+    CreateTeamRequest createTeamRequest =
+        RestFunctions.getJsonBodyAsClass(ctx, CreateTeamRequest.class);
+    processor.createTeam(userData, createTeamRequest);
+    end(ctx.response(), 200);
+  }
```
jackblanc, line 60: typically in other places, we list all of the register methods, then all of the handler methods.

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -1,24 +1,63 @@
...
+import static org.jooq.generated.tables.UsersTeams.USERS_TEAMS;
+
 public class TeamsProcessorImpl implements ITeamsProcessor {
   private final DSLContext db;
 
   public TeamsProcessorImpl(DSLContext db) {
     this.db = db;
   }
 
+  private void checkTeamExists(int teamId) {
```
jackblanc, line 42: public methods listed before private methods in classes

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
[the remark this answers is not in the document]
jackblanc: Monday backlog for SFTT's projects, issues for scaffold-related things

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -35,4 +74,290 @@ public void disbandTeam(JWTData userData, int teamId) {
...
+  @Override
+  public void leaveTeam(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
+    UsersTeamsRecord usersTeamsRecord =
+        db.selectFrom(USERS_TEAMS)
+            .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+            .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+            .fetchOne();
+    if (usersTeamsRecord != null) {
+      if(usersTeamsRecord.getTeamRole() == TeamRole.LEADER) {
```
jackblanc, line 324: we should allow the leader to leave the team if there is > 1 leader

jackblanc (commented):
TeamDoesNotExistException is an empty file
There needs to be test coverage for the processor code

api/src/main/java/com/codeforcommunity/api/ITeamsProcessor.java
oloveluck: Agreed

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
oloveluck: what do you mean by this?

api/src/main/java/com/codeforcommunity/exceptions/LeaderCannotLeaveTeamException.java
oloveluck: No. It would be fine for a leader to leave a team if there is more than one leader. However, there is currently no route for making someone else a team leader without transferring ownership. When a team is created there is only one leader. Given the current API Spec there is no way to get into a state with more than one leader. 
Might want to add this in the future. Would not be a complicated fix.

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
oloveluck: Is there an advantage to this? I personally find this way easier to read, do you prefer the other way?.

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
oloveluck: There is currently no way to reach a state with > 1 leader for a team.

api/src/main/java/com/codeforcommunity/dto/team/TeamDataResponse.java
oloveluck: lowerCamelCase is mostly used. I prefer this too for names of Java Variables. There are some instances when dealing with input from JSON where snake_case is used. I cannot find any instances of snake_case that are unnecessary.

oloveluck:
This PR is too large and should be merged before adding additional code to it. There are a couple TODOs remaining that will not impact frontend development, and I have created an additional branch off of this to make a separate PR. Should have split this up more to avoid this 1000+ line PR...

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
jackblanc: semantically, it should be a plural "users" as this request contains a list of users

api/src/main/java/com/codeforcommunity/dto/team/TransferOwnershipRequest.java
jackblanc: Don't just resolve comments without responding or making a change

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
jackblanc: yeah honestly I like this way better now that I think about it

jackblanc (approved):
Make sure you're responding to comments before resolving them

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "invite_user_request.";
+    List<String> fields = new ArrayList<>();
+
+    if (users == null) {
+      fields.add(fieldName + "users");
+    }
+    // users.forEach(user -> user.validateFields(fieldPrefix));
```
liammoyn, line 34: This block should looks like:
```
if (users == null) {
  fields.add(fieldName + "users");
} else {
  users.forEach(user -> fields.addAll(user.validateFields(fieldPrefix + "users."));
}
```

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +28,143 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+
+  private void registerGetApplicants(Router router) {
+    Route getApplicantsRoute = router.get("/:team_id/applicants");
+    getApplicantsRoute.handler(this::handleGetApplicantsRoute);
+  }
+
+  private void handleGetApplicantsRoute(RoutingContext ctx) {
+    JWTData userData = ctx.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(ctx.request(), "team_id");
+    processor.getApplicants(userData, teamId);
```
liammoyn, line 126: This is a get request, it should be returning data. You need to call the end() method with the object that processor.getApplicants returns

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
liammoyn: Don't resolve conversations without a comment especially when there's no change

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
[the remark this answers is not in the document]
liammoyn: So in this case monday

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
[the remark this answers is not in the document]
liammoyn: Sounds good 👍

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
liammoyn: Add a TODO for this inline if you're not addressing it here

liammoyn (changes requested):
I'm fine with getting this merged and doing followup tickets to address changes but make sure to leave very explicit "TODO" comments in the code where applicable and make sure to create tickets on monday so we don't just forget things

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
oloveluck: dealt with this

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
oloveluck: done

api/src/main/java/com/codeforcommunity/dto/team/TransferOwnershipRequest.java
oloveluck: done

api/src/main/java/com/codeforcommunity/dto/team/InviteUserRequest.java
oloveluck: Disagree with this. If a single email in the list is incorrectly formed, the entire response will fail. The spec says that in this case the incorrectly formed email will be skipped.

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
```diff
@@ -22,12 +31,143 @@ public TeamsRouter(ITeamsProcessor processor) {
...
+
+  private void registerGetApplicants(Router router) {
+    Route getApplicantsRoute = router.get("/:team_id/applicants");
+    getApplicantsRoute.handler(this::handleGetApplicantsRoute);
+  }
+
+  private void handleGetApplicantsRoute(RoutingContext ctx) {
+    JWTData userData = ctx.get("jwt_data");
+    int teamId = RestFunctions.getRequestParameterAsInt(ctx.request(), "team_id");
+    List<UsersTeamDataResponse> response = processor.getApplicants(userData, teamId);
```
liammoyn, line 129: The API response for this route is
```
{
  "applicants": [
    {
      "userId": INT,
      "username": STRING,
    },
    ...
  ]
}
```
Based on that, you need to be returning a single object that has a single field called applicants that is a list of UsersTeamDataResponse objects. With the setup here, you're just returning a JSON Array. Of course currently UsersTeamDataResponse has the wrong fields but we already said we needed to add a TODO for that.

api/src/main/java/com/codeforcommunity/dto/team/UsersTeamDataResponse.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.codeforcommunity.enums.TeamRole;
+
+public class UsersTeamDataResponse {
```
liammoyn, line 5: Either here or in the processor implementation when instantiating this class we need a TODO and a ticket for changing the fields of this class to userId and username as it is in the API specification https://docs.c4cneu.com/sftt/sftt-api-spec/#get-applicants

api/src/main/java/com/codeforcommunity/dto/team/InviteUsersRequest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "invite_user_request.";
+    List<String> fields = new ArrayList<>();
+
+    if (users == null) {
+      fields.add(fieldName + "users");
+    }
+    // If we want to validate fields here: users.forEach(user -> user.validateFields(fieldPrefix));
```
liammoyn, line 34: We do want to validate fields and I already gave you the code block to do that. Lets just add that code instead of a comment

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
liammoyn: I don't think this was addressed. Right now this exception is just saying that the calling user of the route has to be a tean leader, it doesn't say anything about the "newLeaderTeamsRecord" which can also call this to fail so having the one exception for both cases is not correct.

service/src/main/java/com/codeforcommunity/requester/Emailer.java
```diff
@@ -93,4 +93,16 @@ public void sendAccountDeactivatedEmail(String sendToEmail, String sendToName) {
...
+    String filePath = "/emails/InviteEmail.html";
+
+    Map<String, String> templateValues = new HashMap<>();
+    templateValues.put("link", "");
```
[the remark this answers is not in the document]
liammoyn: Still need to add a TODO

liammoyn (changes requested):

service/src/main/java/com/codeforcommunity/requester/Emailer.java
oloveluck: Ticket in Monday

api/src/main/java/com/codeforcommunity/rest/subrouter/TeamsRouter.java
oloveluck: TODO and monday ticket for this

api/src/main/java/com/codeforcommunity/dto/team/UsersTeamDataResponse.java
oloveluck: done

api/src/main/java/com/codeforcommunity/dto/team/InviteUsersRequest.java
oloveluck: I uncommented this for now. See my previous response for this. Validating each of the users causes a problem with this requiirement in the API spec:
"If one of the email addresses is invalid or the user is already on the team that invite will not be send out, the other ones will be and a 200 OK response is returned"
I have made a different ticket to resolve this.

liammoyn: This is different. When we're validating DTOs we're only checking that the JSON is the right _shape_. That means we're checking that each user here has the fields we're expecting and they are not malformed.

What the API is referring to are logical failures which happen when the request we were sent had the right fields, but because of the state of our database (users being on the same team) and/or the execution of our Emailer (email addresses being invalid) the call is invalid. These are checks that happen in the processor because we can't immediately tell from the JSON itself, we need other information.

The API is a little ambiguous about malformed emails (which get checked in the DTO) vs undeliverable emails (which isn't detected until the processor implementation) but I think it's safe to assume that we are okay to return a 400 here in the DTO if any of the emails are malformed.

So, we should have that validateFields check on each of the users as I described above and in the implementation we just have to check if a user is already on the team, and not send them an email if they are.

api/src/main/java/com/codeforcommunity/dto/team/InviteUsersRequest.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "invite_user_request.";
+    List<String> fields = new ArrayList<>();
+
+    if (users == null) {
+      fields.add(fieldName + "users");
+    } else {
+      users.forEach(user -> user.validateFields(fieldPrefix));
```
liammoyn, line 34: Make sure to update the fieldPrefix so it's clear the error happened inside a user object: `validateFields(fieldPrefix + "users.")` and to add the result of each validateFields call to the "fields" object so that it's returned from this method `fields.addAll(user.validateFields(...))`

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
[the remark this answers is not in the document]
liammoyn: https://code4community-team.monday.com/boards/925293934/pulses/1133601799

liammoyn (approved):
Read through all the non-resolved conversations before merging this and be that you can build this locally before merging in.

PR #20, opened 2021-02-21
Opened up map routes to everyone

florisdobber:
Made map routes unprotected since people need to be able see the map on the home page


liammoyn (approved):
Make sure the api specification is updated

PR #23, opened 2021-03-01
Changed delete user route from DELETE to POST request

florisdobber:
No changes to API spec required, already looked like this

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
@@ -11,6 +11,7 @@ import com.codeforcommunity.dto.user.ChangeEmailRequest;
 import com.codeforcommunity.dto.user.ChangePasswordRequest;
 import com.codeforcommunity.dto.user.ChangePrivilegeLevelRequest;
 import com.codeforcommunity.dto.user.ChangeUsernameRequest;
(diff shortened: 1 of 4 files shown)


service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -30,12 +31,17 @@ public ProtectedUserProcessorImpl(DSLContext db, Emailer emailer) {
...
+    if (!Passwords.isExpectedPassword(deleteUserRequest.getPassword(), user.getPasswordHash())) {
+      throw new WrongPasswordException();
+    }
+
     int userId = userData.getUserId();
 
     db.deleteFrom(VERIFICATION_KEYS).where(VERIFICATION_KEYS.USER_ID.eq(userId)).executeAsync();
 
-    UsersRecord user = db.selectFrom(USERS).where(USERS.ID.eq(userId)).fetchOne();
     user.delete();
```
liammoyn, line 45: This is unrelated to the change here but currently I think this would throw an exception in production. We need to a cascading delete where we also delete any rows in the database that have a foreign key relation to this user. Unfortunately, jOOQ doesn't have built in support for this right now so we should either manually delete rows that include a userID in them one-by-one or we could try running a raw SQL query that includes CASCADE. We should create a new ticket for this work so that we don't jam up this PR

liammoyn (approved):

PR #27, opened 2021-03-27
Add getUserTeams

florisdobber:


service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -79,6 +87,34 @@ public UserDataResponse getUserData(JWTData userData) {
...
+    if (user == null) {
+      throw new UserDoesNotExistException(userData.getUserId());
+    }
+
+    Result<Record2<String, Integer>> teams =
+        db.select(TEAMS.TEAM_NAME, TEAMS.ID)
+            .from(USERS_TEAMS)
+            .join(TEAMS)
+            .onKey()
+            .where(USERS_TEAMS.USER_ID.eq(userId))
```
liammoyn, line 104: Don't forget to also filter teams where their status is None or Pending

liammoyn (changes requested):

service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
florisdobber: Good catch, fixed

liammoyn (approved):

PR #30, opened 2021-03-30
Add sql script to extend email field length and update DTO checks

connernilsen:
Some questions about the data types used are in TODO comments :)


florisdobber (approved):
Left one quick comment, rest looks good!

api/src/main/java/com/codeforcommunity/dto/team/AddGoalRequest.java
```diff
@@ -52,13 +52,14 @@ public void setGoal(Integer goal) {
   public List<String> validateFields(String fieldPrefix) throws HandledException {
     String fieldName = fieldPrefix + "add_goal_request.";
     List<String> fields = new ArrayList<>();
-    if (goal == null && goal > 0) {
+    // TODO: is this supposed to be negative?
+    if (goal == null || goal > 0) {
```
florisdobber, line 56: Yeah it is you're right, good catch

api/src/main/java/com/codeforcommunity/dto/team/InviteContact.java
```diff
@@ -35,10 +35,11 @@ public void setEmail(String email) {
   public List<String> validateFields(String fieldPrefix) throws HandledException {
     String fieldName = fieldPrefix + "invite_contact";
     List<String> fields = new ArrayList<>();
-    if (name == null) {
+    // TODO: do we want to keep this field? It doesn't look like it's being used anywhere
```
florisdobber, line 38: We're having some issues with the emailer but eventually this field will be used in the emailer.

PR #32, opened 2021-03-31
add ebconfig to build.zip

jackblanc:


florisdobber (approved):
If this builds works for me

jackblanc:
fun fact @florisdobber this PR changed the only file which does not change the build process 😉

PR #36, opened 2021-04-14
Added stewardship and favorite sites tables

florisdobber:


persist/src/main/resources/db/migration/V3__StewardshipAndFavoriting.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE IF NOT EXISTS stewardship (
+    id          SERIAL      NOT NULL PRIMARY KEY,
+    user_id     INT         NOT NULL,
+    site_id     INT         NOT NULL,
+    date_time   TIMESTAMP   DEFAULT CURRENT_TIMESTAMP,
```
liammoyn, line 5: Kind of weird this is date_time when the request calls it date. Would also be good to have this be more descriptive of when this time is from. Is it when the record was created or when the action was performed?

SofieDunt (commented):
Looks good! Should address the documentation questions before merging

persist/src/main/resources/db/migration/V3__StewardshipAndFavoriting.sql
florisdobber: Changed to performed_on and switched to date type

persist/src/main/resources/db/migration/V3__StewardshipAndFavoriting.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE IF NOT EXISTS stewardship (
+    id              SERIAL      NOT NULL PRIMARY KEY,
+    user_id         INT         NOT NULL,
+    site_id         INT         NOT NULL,
+    performed_on    DATE,
```
SofieDunt, line 5: Should this be `NOT NULL`?

florisdobber: Fixed

SofieDunt (approved):
💯

liammoyn (approved):

PR #40, opened 2021-04-20
More site routes

florisdobber:
Tickets: https://code4community-team.monday.com/boards/925293934/pulses/1185019867
https://code4community-team.monday.com/boards/925293934/pulses/1218648886

I will rename functions and adjust db tables in a separate PR to minimize confusion.

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
@@ -6,9 +6,16 @@ import static org.jooq.generated.Tables.STEWARDSHIP;
 
 import com.codeforcommunity.api.ISiteProcessor;
 import com.codeforcommunity.auth.JWTData;
+import com.codeforcommunity.dto.site.FavoriteSitesResponse;
 import com.codeforcommunity.dto.site.RecordStewardshipRequest;
+import com.codeforcommunity.dto.site.StewardshipActivitiesResponse;
+import com.codeforcommunity.dto.site.StewardshipActivity;
+import com.codeforcommunity.enums.PrivilegeLevel;
+import com.codeforcommunity.exceptions.AuthException;
 import com.codeforcommunity.exceptions.ResourceDoesNotExistException;
 import com.codeforcommunity.exceptions.WrongFavoriteStatusException;
+import java.util.ArrayList;
+import java.util.List;
 import org.jooq.DSLContext;
 import org.jooq.generated.tables.records.FavoriteSitesRecord;
 import org.jooq.generated.tables.records.StewardshipRecord;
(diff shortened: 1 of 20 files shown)


service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -59,6 +66,19 @@ public void unfavoriteSite(JWTData userData, int siteId) {
...
 
+  @Override
+  public FavoriteSitesResponse getFavoriteSites(JWTData userData) {
+    List<FavoriteSitesRecord> favoriteSites =
+        db.selectFrom(FAVORITE_SITES)
+            .where(FAVORITE_SITES.USER_ID.eq(userData.getUserId()))
+            .fetch();
+    List<Integer> favorites = new ArrayList<>();
+
+    favoriteSites.forEach(site -> favorites.add(site.getSiteId()));
```
connernilsen, line 77: Here you can probably do something like 
```java
List<Integer> favorites = db.select(FAVORITE_SITES.SITE_ID)
    .from(FAVORITE_SITES).where(...).fetch();
```
or
```java
List<Integer> favoriteSites =
        db.selectFrom(FAVORITE_SITES)
            .where(FAVORITE_SITES.USER_ID.eq(userData.getUserId()))
            .fetch(FAVORITE_SITES.SITE_ID);
```
to remove the extra logic that loops through each record

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -74,4 +94,45 @@ public void recordStewardship(
...
+              new StewardshipActivity(
+                  record.getId(),
+                  record.getUserId(),
+                  record.getPerformedOn(),
+                  record.getWatered(),
+                  record.getMulched(),
+                  record.getCleaned(),
+                  record.getWeeded());
+          activities.add(stewardshipActivity);
+        });
```
connernilsen, line 134: For this one, you can try `.fetchOneInto(StewardshipActivity.class)`. Depending on how the DTO was set up, you might also need to add a private empty constructor to the class, but we'll look at that if you're getting exceptions.

connernilsen (commented):
Looks good, just a couple questions first.

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
florisdobber: Ah good call, was just copying other code without thinking

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
florisdobber: Can you explain this a little more? fetchOneInto seems to fetch only one record, but I want a list of records.

liammoyn (approved):

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
connernilsen: Oh whoops, I think the method i should have said would be something like 
`.fetchInto(SomeDto.class)`

florisdobber: Is this worth it? Seems to just be confusing things when the current method works

connernilsen (approved):
Looks good

PR #42, opened 2021-04-26
add update and delete site routes

willmt80:
Created update and delete site routes.

Through Postman was able to get the update site route to work. Was not able to get delete to work properly but wanted to get progress up.


api/src/main/java/com/codeforcommunity/dto/site/UpdateSiteRequest.java
```diff
@@ -0,0 +1,440 @@
...
+      String sidewalk_width,
+      Double site_width,
+      Double site_length,
+      String material,
+      Boolean raised_bed,
+      Boolean fence,
+      Boolean trash,
+      Boolean wires,
+      Boolean grate,
+      boolean stump,
```
florisdobber, line 84: Why is this not a `Boolean`?

service/src/main/java/com/codeforcommunity/dataaccess/AuthDatabaseOperations.java
```diff
@@ -96,8 +96,8 @@ public boolean isValidLogin(String email, String pass) {
   }
 
   /**
-   * TODO: Refactor this method to take in a DTO / POJO instance
-   * Creates a new row in the USER table with the given values.
+   * TODO: Refactor this method to take in a DTO / POJO instance Creates a new row in the USER table
```
florisdobber, line 99: This was my doing, sorry

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -1,5 +1,6 @@
 package com.codeforcommunity.processor;
 
+import static org.jooq.generated.Tables.*;
```
florisdobber, line 3: Can you change this back to individual imports? You can disable * imports in the settings as well if it keeps happening

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -93,6 +108,261 @@ public void recordStewardship(
...
+            .orderBy(SITE_ENTRIES.UPDATED_AT.desc())
+            .limit(1)
+            .fetchOne();
+
+    SiteEntriesRecord record = db.newRecord(SITE_ENTRIES);
+
+    record.setUserId(userData.getUserId());
+    record.setSiteId(siteId);
+    record.setUpdatedAt(new Timestamp(System.currentTimeMillis()));
+    if (updateSiteRequest.isTree_present() != null) {
```
florisdobber, line 126: I think the idea is that users will be given a filled in form on the front end with the latest update, and then they can change anything they want. With the implementation you've used it would be impossible to set fields to null if that is the accurate state based on their assessment.

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -93,6 +108,261 @@ public void recordStewardship(
...
+    record.store();
+  }
+
+  @Override
+  public void deleteSite(JWTData userData, int siteId) {
+    isAdminCheck(userData.getPrivilegeLevel());
+    checkSiteExists(siteId);
+
+    db.deleteFrom(SITES).where(SITES.ID.eq(siteId));
+    db.deleteFrom(SITE_ENTRIES).where(SITE_ENTRIES.SITE_ID.eq(siteId));
```
florisdobber, line 363: Sites has a `deleted_at` column, so all you have to do is set that to the current time. No need to actually delete the records from the database.

api/src/main/java/com/codeforcommunity/dto/site/UpdateSiteRequest.java
```diff
@@ -0,0 +1,440 @@
+package com.codeforcommunity.dto.site;
+
+import com.codeforcommunity.dto.ApiDto;
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class UpdateSiteRequest extends ApiDto {
+  private Boolean tree_present;
```
connernilsen, line 9: It looks like we're using camel case in other DTOs. @florisdobber does it matter for this?

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -93,6 +108,261 @@ public void recordStewardship(
...
+    checkSiteExists(siteId);
+
+    SiteEntriesRecord lastEntry =
+        db.selectFrom(SITE_ENTRIES)
+            .where(SITE_ENTRIES.SITE_ID.eq(siteId))
+            .orderBy(SITE_ENTRIES.UPDATED_AT.desc())
+            .limit(1)
+            .fetchOne();
+
+    SiteEntriesRecord record = db.newRecord(SITE_ENTRIES);
```
connernilsen, line 121: If it helps, I think you can also do `db.newRecord(SITE_ENTRIES, lastEntry)` to create a copy. Double check that it does what you want if you're going to use it though.

connernilsen (commented):
Looks good so far!

api/src/main/java/com/codeforcommunity/dto/site/UpdateSiteRequest.java
willmt80: whoops, typo

api/src/main/java/com/codeforcommunity/dto/site/UpdateSiteRequest.java
oloveluck: Yes, this should be camelCase

florisdobber (approved):
Yeah looks good once you fix the other comments! Nice work

api/src/main/java/com/codeforcommunity/dto/site/UpdateSiteRequest.java
```diff
@@ -0,0 +1,440 @@
...
+    this.wires = wires;
+    this.grate = grate;
+    this.stump = stump;
+    this.tree_notes = tree_notes;
+    this.site_notes = site_notes;
+  }
+
+  private UpdateSiteRequest() {}
+
+  public Boolean isTree_present() {
```
rymaju, line 129: Do we want this underscore notation here? We should probably use camel case unless this messes with how its written to JSON

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
rymaju: +1

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -93,6 +109,66 @@ public void recordStewardship(
...
+    record.setStump(updateSiteRequest.isStump());
+    record.setTreeNotes(updateSiteRequest.getTree_notes());
+    record.setSiteNotes(updateSiteRequest.getSite_notes());
+
+    record.store();
+  }
+
+  @Override
+  public void deleteSite(JWTData userData, int siteId) {
+    isAdminCheck(userData.getPrivilegeLevel());
```
rymaju, line 164: ⛏️ IMO we shouldnt have a void method `isAdminCheck` whose only purpose is to throw. I would rather have that logic as a boolean and then throw in the handler itself

rymaju (approved):
Left a couple nitpicks but otherwise looks pretty solid!

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
willmt80: We follow this same pattern for teams and registrations. I can make a ticket to change all of them

PR #46, opened 2021-05-06
made get sites and get stewardship activities public

willmt80:
Split the site router into a public routes and protected routes. Users will be able to view site information and stewardship activities on tree pages even if they aren't logged in.


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -0,0 +1,247 @@
...
+    this.db = db;
+  }
+
+  private void checkSiteExists(int siteId) {
+    if (!db.fetchExists(db.selectFrom(SITES).where(SITES.ID.eq(siteId)))) {
+      throw new ResourceDoesNotExistException(siteId, "Site");
+    }
+  }
+
+  private Boolean isAlreadyAdopted(int userId, int siteId) {
```
connernilsen, line 40: Just for clarification, a site can be adopted by multiple users? If that's true, then maybe change the method name here to `isAlreadyAdoptedByUser()` to make it more clear.

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  @Override
+  public void unadoptSite(JWTData userData, int siteId) {
+    checkSiteExists(siteId);
+    if (!isAlreadyAdopted(userData.getUserId(), siteId)) {
+      throw new WrongAdoptionStatusException(false);
+    }
+
+    db.deleteFrom(ADOPTED_SITES)
```
connernilsen, line 78: I haven't seen the schema for this db in a while, but just to double check, are we not doing soft deletes on this table?

connernilsen (commented):
Good work! I just have a few questions before I approve

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
willmt80: I'm not sure, @florisdobber @oloveluck do either of you have thoughts on this?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
willmt80: No, a site can only be adopted by 1 person. Added the isAlreadyAdoptedByUser method for unadopting.

connernilsen (approved):
I think it looks good 👍

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -0,0 +1,252 @@
...
+            .where(ADOPTED_SITES.USER_ID.eq(userId))
+            .and(ADOPTED_SITES.SITE_ID.eq(siteId)));
+  }
+
+  /**
+   * Throws an exception if the user is not an admin or super admin.
+   *
+   * @param level the privilege level of the user calling the route
+   */
+  void isAdminCheck(PrivilegeLevel level) {
```
rymaju, line 56: ❓ Should this belong in its own utility file?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -0,0 +1,252 @@
...
+    return db.fetchExists(
+        db.selectFrom(ADOPTED_SITES)
+            .where(ADOPTED_SITES.USER_ID.eq(userId))
+            .and(ADOPTED_SITES.SITE_ID.eq(siteId)));
+  }
+
+  /**
+   * Throws an exception if the user is not an admin or super admin.
+   *
+   * @param level the privilege level of the user calling the route
```
rymaju, line 54: ⛏️ Should have a `@throws` javadoc

rymaju (commented):
Overall looks good to me 👍

florisdobber (approved):

PR #51, opened 2021-05-14
Added target user exists check to change privilege level

florisdobber:
https://code4community-team.monday.com/boards/925293934/pulses/1270290066


service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
```diff
@@ -174,7 +175,14 @@ public void changePrivilegeLevel(
...
     String targetEmail = changePrivilegeLevelRequest.getTargetUserEmail();
+
+    // check if target user exists
+    if (!db.fetchExists(db.selectFrom(USERS).where(USERS.EMAIL.eq(targetEmail)))) {
+      throw new UserDoesNotExistException(targetEmail);
+    }
+
     UsersRecord user = db.selectFrom(USERS).where(USERS.EMAIL.eq(targetEmail)).fetchOne();
 
     userRecordExistsCheck(user, new UserDoesNotExistException(targetEmail));
```
SofieDunt, line 188: ❓ What does this do?

SofieDunt (approved):
Looks good to me!

service/src/main/java/com/codeforcommunity/processor/ProtectedUserProcessorImpl.java
florisdobber: You're right, does the same thing. Removed second check as it relies on the db call that was causing the error.

PR #53, opened 2021-05-14
Fixed goal value in getTeam

florisdobber:
https://code4community-team.monday.com/boards/925293934/pulses/1167889035


SofieDunt (approved):
👍 Nice catch!

PR #55, opened 2021-05-17
Added planting date to db and geojson route

florisdobber:


service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
```diff
@@ -137,7 +148,8 @@ private SiteFeature siteFeatureFromRecord(
             sitesRecord.value4(),
             sitesRecord.value5(),
             sitesRecord.value6(),
-            sitesRecord.value7());
+            sitesRecord.value7(),
+            sitesRecord.value10());
```
willmt80, line 152: ❓ Just for consistency should we switch around the order in the record so that the Date is value8?

willmt80 (approved):
Looks good! Pretty straightforward. Left one thought but it's a pretty minor nitpick

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
florisdobber: Done

SofieDunt (approved):
Looks good to me!

PR #56, opened 2021-05-23
changed from statement for entry usernames and made mergeSorted iterative

willmt80:
Resolved an error where nullUserRecords was trying to select from the users table even though the Users table was not in the FROM statement. Now that table is selecting from Entry_Usernames.

Also resolved a stackoverflow error in mergeSorted by changing the function from recursive to iterative.


rymaju (approved):
Looks solid 💯

florisdobber (approved):
💯

PR #57, opened 2021-05-25
Fix record stewardship

SofieDunt:
Fixed [issue](https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1323567094) where stewardship activities could not be recorded by setting the stewardship record's siteId.


florisdobber (approved):
Good catch!

willmt80 (approved):
Looks good! Do you think we want to enforce that a user must adopt a site before they can perform stewardship activities?

PR #61, opened 2021-06-12
removed username from siteGeoJSON

willmt80:
Changed the siteGeoJson query to not look for username. Not a huge performance increase but enough to be significant, particularly on the longest response

old:
current-latency

new:
new-latency

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
+++ b/api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
@@ -30,7 +30,6 @@ public class SiteFeatureProperties {
     return id;
   }
 
-  @JsonProperty("tree_present")
   public Boolean getTreePresent() {
     return treePresent;
   }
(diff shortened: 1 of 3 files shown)


rymaju (approved):
LGTM

Make sure to run mvn spotless:apply for formatting

api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
```diff
@@ -45,22 +42,12 @@ public Boolean getTreePresent() {
     return treePresent;
   }
 
-  public Double getDiameter() {
+  public Double diameter() {
```
SofieDunt, line 45: What's the naming convention for this? (diameter vs getDiameter)

SofieDunt (approved):
👍

api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
willmt80: Removed this from feature

PR #67, opened 2021-07-09
Unique site entries on map

willmt80:
Edited the map site query in order to prevent repeating sites from appearing on the map.

Instead of checking if a timestamp is in a subquery, the query uses a series of joins.

The latency for the longest request is slightly longer than before but the mean latency is lower.

latency

Here is an example of the old output on the frontend

image

And the same spot with the new query

no-overlap

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
+++ b/api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
@@ -1,6 +1,5 @@
 package com.codeforcommunity.dto.map;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import java.sql.Date;
 
 public class SiteFeatureProperties {
--- a/service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
@@ -21,22 +21,21 @@ import com.codeforcommunity.dto.map.NeighborhoodGeoResponse;
 import com.codeforcommunity.dto.map.SiteFeature;
 import com.codeforcommunity.dto.map.SiteFeatureProperties;
 import com.codeforcommunity.dto.map.SiteGeoResponse;
-import com.codeforcommunity.dto.map.TimeStamp;
 import com.codeforcommunity.enums.ReservationAction;
 import com.codeforcommunity.logger.SLogger;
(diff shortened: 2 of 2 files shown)


api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
```diff
@@ -1,6 +1,5 @@
 package com.codeforcommunity.dto.map;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
```
rymaju, line 3: Assuming this is an unused import?

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
```diff
@@ -171,6 +176,42 @@ public SiteGeoResponse getSiteGeoJson() {
     if (!SiteGeoResponseCache.isExpired()) {
       return SiteGeoResponseCache.getResponse();
     }
+
+    Field<Timestamp> maxDate = max(SITE_ENTRIES.UPDATED_AT).as("maxDate");
+
+    Table<
```
rymaju, line 182: Is it just me or is the formatting for this really odd?

Have you run `mvn spotless:apply`?

rymaju (commented):
Goods but the formatting is weird, did you run the formatter? 🍕

api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
willmt80: This was automatically removed with mvn spotless:apply

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
willmt80: Yeah I ran it, idk why the formatting is so funky

api/src/main/java/com/codeforcommunity/dto/map/SiteFeatureProperties.java
willmt80: Doesn't look like JsonProperty is used in that file so I think it's good

SofieDunt (approved):
Looks good! Are we still trying to get caching to work in prod?

rymaju (approved):
funky formatting but hey awesome work 💯

PR #71, opened 2021-07-12
edit configuration to include email

willmt80:
I'm not totally positive if this fixed everything but I copied the configuration from the llb build files.

Not sure how to access the secrets in ElasticBeanstalk so there may need to be an edit there too


rymaju (approved):
Huh yeah im ok with merging this but did you reach out to whoever set this up initially? Maybe we should set up a meeting to better understand how this works.

PR #73, opened 2021-07-17
Added list of team members to getTeam request

willmt80:
Part of this ticket: https://code4community-team.monday.com/boards/925293934/views/18904377/pulses/1133710526

The team page needs to display which members are on that team. Added a query and types to get a list of user information for each member of a team.

request still works quickly when called in postman so join is not too inefficient.


api/src/main/java/com/codeforcommunity/dto/team/TeamMembersResponse.java
```diff
@@ -0,0 +1,27 @@
+package com.codeforcommunity.dto.team;
+
+import com.codeforcommunity.enums.TeamRole;
+
+public class TeamMembersResponse {
```
SofieDunt, line 5: 🔥

SofieDunt (commented):
Looks good to me! I think we should consider removing the getMembers route or having it return a list of TeamMembersResponse as part of this ticket

willmt80:
> Looks good to me! I think we should consider removing the getMembers route or having it return a list of TeamMembersResponse as part of this ticket

I think whether or not we keep the route kind of depends on how we want to allow team leaders to approve new members. I'm not entirely sure what the point of that route is since it doesn't have the usernames

rymaju:
@chromium-52 Will this introduce a breaking change? Close or merge?

chromium-52:
> @chromium-52 Will this introduce a breaking change? Close or merge?

@rymaju Sadly, I don't have any context on this PR (@SofieDunt do you happen to remember this?). Otherwise, I think we can merge. The route isn't used anywhere (and it may never be) so it won't break anything

rymaju:
Ok, merging

PR #75, opened 2021-09-08
Wt.dev docs update

willmt80:
Small updates to backend to match dev docs

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
@@ -220,8 +220,20 @@ public class TeamsProcessorImpl implements ITeamsProcessor {
   }
 
   @Override
-  public UsersResponse getApplicants(int teamId) {
-    return getUsers(teamId, TeamRole.PENDING);
+  public UsersResponse getApplicants(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
+    UsersTeamsRecord userTeamsRecord =
+            db.selectFrom(USERS_TEAMS)
+                    .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+                    .and(USERS_TEAMS.TEAM_ID.eq(teamId))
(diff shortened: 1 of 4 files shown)


SofieDunt (approved):
Thanks for doing this! 💯

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -76,6 +76,8 @@ public TeamsProcessorImpl(DSLContext db) {
 
   @Override
   public void disbandTeam(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
```
SofieDunt, line 79: 💯

service/src/main/java/com/codeforcommunity/processor/TeamsProcessorImpl.java
```diff
@@ -218,8 +220,20 @@ public void inviteUser(JWTData userData, InviteUsersRequest inviteUserRequest, i
...
-  public UsersResponse getApplicants(int teamId) {
-    return getUsers(teamId, TeamRole.PENDING);
+  public UsersResponse getApplicants(JWTData userData, int teamId) {
+    checkTeamExists(teamId);
+    UsersTeamsRecord userTeamsRecord =
+            db.selectFrom(USERS_TEAMS)
+                    .where(USERS_TEAMS.USER_ID.eq(userData.getUserId()))
+                    .and(USERS_TEAMS.TEAM_ID.eq(teamId))
+                    .and(USERS_TEAMS.TEAM_ROLE.eq(TeamRole.LEADER))
+                    .fetchOne();
```
SofieDunt, line 230: ❓ This is a check that the user is team leader, right? Can this logic also be used for `disbandTeam`?

willmt80: Yes that's true, just abstracted that logic out because it's used all over the place.

rymaju (approved):
LGTM

maxsebso (approved):
LGTM

PR #77, opened 2021-09-20
Wt.neighborhoods to site

willmt80:
Clickup ticket: https://app.clickup.com/t/1dk47gz

Added neighborhood_id to the site table with a migration script. Ran mvn clean install to generate new site record files. Updated the getSite request to include the neighborhood name.

neighborhood


service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -133,6 +134,12 @@ public GetSiteResponse getSite(int siteId) {
       throw new ResourceDoesNotExistException(siteId, "site");
     }
 
+    String neighborhood =
+        db.selectFrom(NEIGHBORHOODS)
+            .where(NEIGHBORHOODS.ID.eq(sitesRecord.getNeighborhoodId()))
+            .fetchOne()
```
connernilsen, line 140: This can't be null, right? Just want to double check

connernilsen (approved):
Looks great!

vrushalitarte (approved):
looks good to me!

chromium-52 (approved):
Looks great!

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
willmt80: Correct, this is a required field

PR #80, opened 2021-09-29
Added condition such that species is returned when common name is null

willmt80:
https://app.clickup.com/t/1fxr8vj

Return the genus and species of a tree if the site does not have a common name (fairly uncommon occurrence)

Works for when either the genus or species is null as well, just makes the select statement pretty difficult to read.

Load Test Comparison:

Master:
old_site_latency

This PR:
load_sites_new

The requests on master are faster than with the new query but not by so much that it will bring the site to a crawl, importantly the amount of data the request returns remains the same, potentially more slowdown if we pass all 3 columns to the frontend instead of combining them.

Apparently string concatenation with jooq is bad practice, something to consider: https://blog.jooq.org/never-concatenate-strings-with-jooq/


service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
```diff
@@ -206,17 +210,43 @@ public SiteGeoResponse getSiteGeoJson() {
...
+                                newEntries.field(SITE_ENTRIES.SPECIES))
+                            .otherwise( // common name and species are empty
+                                when(
+                                        newEntries.field(SITE_ENTRIES.SPECIES).eq(""),
+                                        newEntries.field(SITE_ENTRIES.GENUS))
+                                    .otherwise( // just common name is empty
+                                        concat(
+                                            newEntries.field(SITE_ENTRIES.GENUS).concat(" "),
+                                            newEntries.field(SITE_ENTRIES.SPECIES))))))
+            .otherwise(newEntries.field(SITE_ENTRIES.COMMON_NAME)); // common name is not empty
```
SofieDunt, line 243: 👍  nice commenting

SofieDunt (approved):
Looks good to me!

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
vrushalitarte: ^yes very helpful!

vrushalitarte (approved):

PR #81, opened 2021-10-03
updated site import to account for neighborhoods

willmt80:
https://app.clickup.com/t/1jt6z9c

Recently made changes to the backend script and production database that add neighborhood IDs to the site table. Updating the site import script to ensure that there is not a discrepancy between the data in production and the data on people's local databases.

Changes include:

- Adding neighborhood_id to the import script
- Updated the site.json with all of the information currently in production
- Modified the length of the species column to account for values with length >30 (production already had this change but it was not in a migration script)
- Updated add site route to include neighborhood_id


rymaju:
Dear god the diff is huge. Can we store this json in S3 instead of in github?

PR #84, opened 2021-10-18
Wt.add sites

willmt80:
https://app.clickup.com/t/1dk48g9

Creates a route where admins can create more than one site at a time. Uses logic already defined to add individual sites. 

Also modified some of the constraints on add site and update site such as defaulting boolean values to false and making block_id an optional field.

Originally attempted to handle csv parsing and processing on the server, however given the small size of the the expected csvs and familiarity with csv parsing in Typescript, decided to handle csv processing on the frontend.


chromium-52 (approved):
Looks great!

PR #88, opened 2021-11-01
Update naming conventions

chromium-52:
Changed:
Underscores in method names and variable names of several files

Didn't change:
Underscores in constant names (e.g. DEFAULT_PREVIOUS_DAYS), comments (e.g. line 42 in [ProtectedSiteProcessorImpl.java](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/blob/master/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java)), route names (e.g. line 64 of [AuthRouter.java](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/blob/master/api/src/main/java/com/codeforcommunity/rest/subrouter/AuthRouter.java)), string constants (e.g. 'forgot_password_request' in line 27 of [ForgotPasswordRequest.java](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/blob/master/api/src/main/java/com/codeforcommunity/dto/auth/ForgotPasswordRequest.java)), and non-.java files (.html, .xml, .sql, etc.)


vrushalitarte (approved):
not sure why the comments in ProtectedSiteProcessorImpl.java use underscores when the variables don't, I'm guessing it's because they refer to the database and so it makes sense to keep those as is -- if there isn't a particular reason I think those should be changed to match the camelCase format. looks good otherwise, thanks for doing this!

SofieDunt (commented):
This seems a lot better to me. Did you check that there's no issue with the routes? I think I've had some issues with the JSON conversion in the past.

chromium-52:
@vrushalitarte Ah I see. I also removed the underscores in comments.

@SofieDunt I hope the routes don't mind my changes too much but I tested routes like adding and editing sites and they still seem to work. I renamed the methods using the refactoring feature instead of changing their names individually so I think they should still work? What kind of issues were you having?

SofieDunt (approved):
🔥 Seems like it was just something weird on my end - great work!

PR #89, opened 2021-11-16
added check to ensure user has adopted a particular site for recording stewardship

willmt80:
Ticket: https://app.clickup.com/t/1dk490u

Ensures that the user has adopted the given site before recording a stewardship activity. Returns a 400 exception if they are not the adopter.

The check if user has adopted site method and exception already existed, just added them to the record stewardship method


vrushalitarte (approved):
lgtm!

SofieDunt (approved):
💯

PR #93, opened 2022-01-10
Add CSV adoption and stewardship report routes

chromium-52:
[Clickup ticket](https://app.clickup.com/t/1q8byxp)

Added routes that return adoption and stewardship reports as a CSV.

Tested the new routes with various `previousDays` parameters, such as valid values (10), zeros, negative values (e.g., -1), decimal values (e.g., 2.4), non-numeric values (e.g. abc), and omitting the parameter. Tested that the existing non-CSV routes still work.


service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -54,18 +63,55 @@ public GetAdoptionReportResponse getAdoptionReport(JWTData userData) {
...
+        LocalDate startLocalDate = previousDays == MISSING_PREVIOUS_DAYS
+                ? LocalDate.ofEpochDay(0) : LocalDate.now().minusDays(previousDays);
+        long startMilli = startLocalDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli();
+
+        List<AdoptedSite> adoptedSites = queryAdoptedSites(startMilli);
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("Site ID, Address, Name, Email, Date Adopted, Activity Count, Neighborhood\n");
+        for (AdoptedSite site : adoptedSites) {
+            builder.append(site.getSiteId() + ", " + site.getAddress() + ", " + site.getName() + ", " + site.getEmail() +
```
SofieDunt, line 101: ⛏️ Prefer chained appends in for loop

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -54,18 +63,55 @@ public GetAdoptionReportResponse getAdoptionReport(JWTData userData) {
...
+        return builder.toString();
+    }
+
+    /**
+     * Query stewardship activities that have been performed at or after the given date.
+     *
+     * @param startDate stewardship activities performed after this date are included in the returned list
+     * @return list of stewardship activities that have been performed at or after the given date
+     */
+    private List<Stewardship> queryStewardships(Date startDate) {
```
SofieDunt, line 114: 💯

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -76,9 +122,43 @@ public GetStewardshipReportResponse getStewardshipReport(JWTData userData) {
...
 
+        return stewardships;
+    }
+
+    @Override
+    public GetStewardshipReportResponse getStewardshipReport(JWTData userData) {
+        isAdminCheck(userData.getPrivilegeLevel());
+
+        // get all stewardships
+        List<Stewardship> stewardships = queryStewardships(new Date(0));
```
SofieDunt, line 137: 👍

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -76,9 +122,43 @@ public GetStewardshipReportResponse getStewardshipReport(JWTData userData) {
...
+
         return new GetStewardshipReportResponse(stewardships);
     }
+
+    @Override
+    public String getStewardshipReportCSV(JWTData userData, GetReportCSVRequest getReportCSVRequest) {
```
SofieDunt, line 143: Same comments as above (startLocalDate and StringBuilder)

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedReportRouter.java
```diff
@@ -55,4 +83,43 @@ private void handleGetStewardshipReportRoute(RoutingContext ctx) {
...
+     */
+    private Integer getValidatedPreviousDaysParam(RoutingContext ctx) {
+        Optional<Integer> maybePreviousDays =
+                RestFunctions.getOptionalQueryParam(ctx, PREVIOUS_DAYS_QUERY_PARAM_NAME, Integer::parseInt);
+        Integer previousDays = maybePreviousDays.orElse(null);
+
+        // validate user input
+        if (previousDays == null || System.currentTimeMillis() - TimeUnit.DAYS.toMillis(previousDays) < 0) {
+            previousDays = DEFAULT_PREVIOUS_DAYS;
+        } else if (previousDays < 0) {
```
SofieDunt, line 119: ❓ When is this true but `System.currentTimeMillis() - TimeUnit.DAYS.toMillis(previousDays) < 0` isn't? Aren't they both checks for the future?

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedReportRouter.java
```diff
@@ -55,4 +83,43 @@ private void handleGetStewardshipReportRoute(RoutingContext ctx) {
...
+     * Return the validated previousDays query parameter from the request URL. The parameter must be an integer greater
+     * than or equal to 0. If the parameter is not given, a value of DEFAULT_PREVIOUS_DAYS is returned.
+     *
+     * @param ctx routing context
+     * @return validated previousDays parameter
+     */
+    private Integer getValidatedPreviousDaysParam(RoutingContext ctx) {
+        Optional<Integer> maybePreviousDays =
+                RestFunctions.getOptionalQueryParam(ctx, PREVIOUS_DAYS_QUERY_PARAM_NAME, Integer::parseInt);
+        Integer previousDays = maybePreviousDays.orElse(null);
```
SofieDunt, line 114: I think you can just do `.orElse(daysSinceJan1970)` here. If `maybePreviousDays` is null, then that's the desired behavior and if it's in the future, it's valid for the query to return an empty report since that's technically the report since some future date.

SofieDunt (commented):
Good work on this! Left a couple questions and a few suggestions for simplicity.

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -54,18 +63,55 @@ public GetAdoptionReportResponse getAdoptionReport(JWTData userData) {
...
+    public String getAdoptionReportCSV(JWTData userData, GetReportCSVRequest getReportCSVRequest) {
         isAdminCheck(userData.getPrivilegeLevel());
 
+        Integer previousDays = getReportCSVRequest.getPreviousDays();
+        // if previousDays parameter is not given, all adopted sites are returned
+        LocalDate startLocalDate = previousDays == MISSING_PREVIOUS_DAYS
+                ? LocalDate.ofEpochDay(0) : LocalDate.now().minusDays(previousDays);
+        long startMilli = startLocalDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli();
+
+        List<AdoptedSite> adoptedSites = queryAdoptedSites(startMilli);
```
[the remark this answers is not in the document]
chromium-52: Fixed! :)

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedReportRouter.java
chromium-52: `previousDays < 0` was true when the time was in the future whereas `System.currentTimeMillis() - TimeUnit.DAYS.toMillis(previousDays) < 0` checked if the date wasn't before the Unix epoch. Without `System.currentTimeMillis() - TimeUnit.DAYS.toMillis(previousDays) < 0`, passing in a number like 10000000 as `previousDays` threw a 500 Internal Server error because the date is out or range.

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -33,16 +38,68 @@ public ProtectedReportProcessorImpl(DSLContext db) {
...
     }
 
+    /**
+     * Converts the previousDays parameter into a Date object. The earliest possible date returned is January 1, 1970
+     * (Unix epoch).
+     *
+     * @param getReportCSVRequest CSV report route request DTO
+     * @return a Date object which is previousDays days before the current date
+     */
+    private Date getStartDate(GetReportCSVRequest getReportCSVRequest) {
```
SofieDunt, line 54: 👍

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
```diff
@@ -54,18 +111,53 @@ public GetAdoptionReportResponse getAdoptionReport(JWTData userData) {
                 .on(ADOPTED_SITES.SITE_ID.eq(STEWARDSHIP.SITE_ID))
                 .leftJoin(NEIGHBORHOODS)
                 .on(SITES.NEIGHBORHOOD_ID.eq(NEIGHBORHOODS.ID))
+                .where(ADOPTED_SITES.DATE_ADOPTED.ge(startDate))
+                .or(TimeUnit.MILLISECONDS.toDays(startDate.getTime()) == 0)
```
SofieDunt, line 115: ❓ Do you still need the `.or`? The stewardship query doesn't have it and passing `Date(0)` should work as expected, right?

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedReportRouter.java
SofieDunt: Ah, good to know, nice case coverage!

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
chromium-52: Without the `or`, the non-CSV route and the CSV route when the `previousDays` parameter is not given (in other words, cases where we want to get *all* adopted sites) don't work as expected for me. In these cases, sites whose `date_adopted` values are null aren't included in the response. This is because, for sites with null `date_adopted` values, the `where` clause on line 114 becomes `where(null.ge(startDate))` and this is always false, regardless of `startDate`.

The reason why the stewardship query doesn't have the `or` is because there seems to be a constraint that the `performed_on` value of a stewardship activity can't be null.

SofieDunt: Forgot that some of the values were null. This is good, but we should consider updating the data with the dates of the weeks the trees were adopted.

SofieDunt (approved):
💯 Looks good to me!

vrushalitarte (approved):

service/src/main/java/com/codeforcommunity/processor/ProtectedReportProcessorImpl.java
chromium-52: Do you mean updating the adopted sites table so that the `DATE_ADOPTED` field can't be null like how it's done with the stewardship table (and maybe backfill some of the adopted sites whose `DATE_ADOPTED` field is already null)?

PR #94, opened 2022-02-02
Fixed the GetSiteResponse object

Tvcz:
Fixed the GetSiteResponse object to use a neighborhood's ID rather than its name, and changed SiteProcessorImpl to pass the correct value to the GetSiteResponse constructor. @jhylisa666


SofieDunt (approved):
Looks great! 🔥

jhylisa666 (commented):
Looks good to me

jhylisa666 (approved):

PR #97, opened 2022-02-09
Jh.lj Force Unadopt Route for Admins

Tvcz:
@jhylisa666 and I worked on a new route that allows admins to force unadopt the sites of standard users and allows super admins to force unadopt the sites of standard users, admins, and other super admins.

In the course of modifying the ProtectedSiteProcessor, we also added a new utility method `isAdmin(PrivilegeLevel)` which returns a boolean indicating whether a user with the given privilege level is an admin or super admin.

Route is accessible via:
`POST api/v1/protected/sites/:site_id/force_unadopt`

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
@@ -131,12 +131,12 @@ public class ProtectedSiteProcessorImpl implements IProtectedSiteProcessor {
       throw new WrongAdoptionStatusException(false); // wrong error?
     }
 
-    AdoptedSitesRecord result = db.selectFrom(ADOPTED_SITES)
+    AdoptedSitesRecord adoptedSite = db.selectFrom(ADOPTED_SITES)
             .where(ADOPTED_SITES.SITE_ID.eq(siteId))
             .fetchInto(AdoptedSitesRecord.class)
             .get(0);
 
-    Integer adopterId = result.getUserId();
+    Integer adopterId = adoptedSite.getUserId();
 
     UsersRecord adopter = db.selectFrom(USERS)
             .where(USERS.ID.eq(adopterId))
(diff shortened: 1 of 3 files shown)


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -119,6 +131,34 @@ public void unadoptSite(JWTData userData, int siteId) {
         .execute();
   }
 
+  @Override
+  public void forceUnadoptSite(JWTData userData, int siteId) {
+    isAdminCheck(userData.getPrivilegeLevel());
+    checkSiteExists(siteId);
+    if (!isAlreadyAdopted(siteId)) {
+      throw new WrongAdoptionStatusException(false); // wrong error?
```
SofieDunt, line 139: 👍 This is a good exception to use, it'll return with message "Site is not marked as adopted", which makes sense

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -119,6 +131,34 @@ public void unadoptSite(JWTData userData, int siteId) {
...
+            .fetchInto(AdoptedSitesRecord.class)
+            .get(0);
+
+    Integer adopterId = adoptedSite.getUserId();
+
+    UsersRecord adopter = db.selectFrom(USERS)
+            .where(USERS.ID.eq(adopterId))
+            .fetchOne();
+
+    if(isAdmin(adopter.getPrivilegeLevel()) && !(userData.getPrivilegeLevel() == (PrivilegeLevel.SUPER_ADMIN))) {
```
SofieDunt, line 153: ⛏️ Super small – this is using `==` for enum equality, but above it uses `.equals`. Both are correct, but we should keep it consistent

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -119,6 +131,34 @@ public void unadoptSite(JWTData userData, int siteId) {
         .execute();
   }
 
+  @Override
+  public void forceUnadoptSite(JWTData userData, int siteId) {
+    isAdminCheck(userData.getPrivilegeLevel());
+    checkSiteExists(siteId);
```
SofieDunt, line 137: 👍

SofieDunt (approved):
💯 Looks good to me! Great work on this!

chromium-52 (approved):
Looks great! Could you just be able to add this route to the API spec page at https://github.com/Code-4-Community/c4c-dev-docs/blob/master/docs/sftt/sftt-api-spec.md?

PR #102, opened 2022-02-16
Created migration script for adding parent account table

Tvcz:
Worked with @chromium-52 to add migration script `V3_11__AddParentAccount.sql` that creates a new table `parent_accounts` that links a parent user account with a child user account. 

[ClickUp Ticket](https://app.clickup.com/t/27pkzc1)

### DIFF
--- a/persist/src/main/resources/db/migration/V3_11__AddParentAccount.sql
+++ b/persist/src/main/resources/db/migration/V3_11__AddParentAccount.sql
@@ -1,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS parent_accounts (
+    parent_id     INT     NOT NULL,
+    child_id      INT     NOT NULL,
+
+    CONSTRAINT parent_accounts_parent_id_fk FOREIGN KEY (parent_id) REFERENCES users (id),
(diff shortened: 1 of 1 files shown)


persist/src/main/resources/db/migration/V3_11__AddParentAccount.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS parent_accounts (
+    parent_id     INT     NOT NULL,
+    child_id      INT     NOT NULL,
+
```
SofieDunt, line 4: 🔧 We should have a primary key here too, can be just `PRIMARY KEY (parent_id, child_id)`

rymaju: Not a SQL expert but does this not deserve its own autoincrementing id?

SofieDunt: It could, but the `parent_id` and `child_id` combination should be unique and is more meaningful than an autoincrementing id

SofieDunt (approved):
LGTM 👍

chromium-52 (approved):

PR #104, opened 2022-03-06
Edit canopy coverage

aschimmichanga:
Added full functionality to be able to edit the canopy coverage of a specified neighborhood and check validity of given canopy coverage and neighborhood ID.
Tested with Postman.


chromium-52:
What I usually do when I end up in situations like this is a [git rebase](https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request) (I don't know how to do this with a GUI).

In general, I think it's a good idea to switch to the `master` branch and do a `git pull` right before creating the feature branch to prevent something like this.

SofieDunt:
Closed by #112

PR #105, opened 2022-03-20
Route to name most recent site entry for an adopted site

Tvcz:
Created a route for renaming the most recent site entry that exists for a given site id. The site must be adopted by the calling user.

Route accessible via:
`POST api/v1/protected/sites/:site_id/name_entry`

Also created a LinkedResourceDoesNotExist exception in order to give more information in the case of a resource not existing that on its own does not have a meaningful id. The exception responds with a `400 BAD REQUEST` status. An example of the error message is shown below:

Screenshot 2022-03-20 144916

[ClickUp Ticket](https://app.clickup.com/t/22gq4y3)

### DIFF
--- a/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
+++ b/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
@@ -48,6 +48,9 @@ public interface IProtectedSiteProcessor {
   /** Removes the given stewardship activity */
   void deleteStewardship(JWTData userData, int activityId);
 
-  /** Removes the record in the adopted sites table linking the site to its current adopter */
+  /**
+   * Renames the latest site entry of the site with the given siteId
+   * using the new name specified in the nameSiteEntryRequest
+   */
   void nameSiteEntry(JWTData userData, int siteId, NameSiteEntryRequest nameSiteEntryRequest);
 }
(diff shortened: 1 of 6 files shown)


rymaju (approved):
LGTM 🛳⛴🚢🏴‍☠️🏴‍☠️🏴‍☠️🚀🚀🚀🛸

api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
```diff
@@ -46,4 +47,7 @@ void recordStewardship(
 
   /** Removes the given stewardship activity */
   void deleteStewardship(JWTData userData, int activityId);
+
+  /** Removes the record in the adopted sites table linking the site to its current adopter */
```
SofieDunt, line 51: ❓ Is this correct?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -374,4 +378,28 @@ public void deleteStewardship(JWTData userData, int activityId) {
...
+      throw new AuthException("User is not the site's adopter.");
+    }
+
+    SiteEntriesRecord siteEntry = db.selectFrom(SITE_ENTRIES)
+        .where(SITE_ENTRIES.SITE_ID.eq(siteId))
+        .orderBy(SITE_ENTRIES.UPDATED_AT)
+        .limit(1)
+        .fetchOne();
+
+    if (siteEntry == null) {
```
SofieDunt, line 394: 👍 Good check

api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
Tvcz: Nope, thanks. Should be fixed now. :+1:

PR #107, opened 2022-04-03
Email inactive users

chromium-52:
Clickup: https://app.clickup.com/t/1hwf0b3

Create a route that checks every 21 days if an adopted site hasn't been taken care of by its adopter within that period (last 21 days). If so, send an email to the adopter about their site being inactive. All of the inactive sites for a given user will be aggregated into one email so that even if a user has multiple inactive sites, they will only receive one email every at most 21 days. The user will continue to receive an email every 21 days until they record a stewardship activity or unadopt the site.

(Updated) example email: 
image
(Link takes user to tree page).


rymaju (commented):
Will this spam users who havent updated their site in a while?

chromium-52:
> Will this spam users who havent updated their site in a while?

I changed it so that users will only receive at most 1 email in at least every 21 days even if they have multiple inactive adopted sites 👍

SofieDunt (approved):
Makes sense to me - email looks good!

zhangcry (approved):

PR #110, opened 2022-04-09
Specify site ID in addSite route

chromium-52:
Explicitly specify the ID of the `siteEntriesRecord` when creating one as it was giving me a duplicate primary key error without it when I was testing the add site form in the frontend (similar to line 206 of `ProtectedSiteProcessorImpl.java` with `sitesRecord`).


Tvcz (approved):
lgtm. Weird though, does this table not have automatic iteration for the primary key?

SofieDunt (approved):
👍

PR #115, opened 2022-08-07
Rename getNeighborhood()

chromium-52:
Rename `getNeighborhood()` in `GetSiteResponse.java` to `getNeighborhoodId()` to match the fix made in #94 to change the neighborhood field returned by the getSite route to neighborhood ID. This also fixes the neighborhood dropdown not showing the neighborhood name correctly.

before
after


zhangcry (approved):

Tvcz (approved):
lgtm. Though given the egregious length it should probably be split into two PRs ;)

chromium-52:
> lgtm. Though given the egregious length it should probably be split into two PRs ;)

That's an oversight on my end, I definitely should've made a separate PR for each letter 😞

PR #116, opened 2022-10-22
Child Account Functionality

Tvcz:
Worked with @mayazeldin to create a new route to handle parents adopting sites for a specified child and a new route to handle parents recording stewardship activities for a specified child.

Route to handle parent adopt site for child:
`POST api/v1/protected/sites/:site_id/parent_adopt`

Route to handle parent record stewardship for child:
`POST api/v1/protected/sites/:site_id/parent_record_stewardship`

Created additional helper methods (`checkParent`, `isParent`) for the `ProtectedSiteProcessor` for determining whether a user is the parent of a given child.

[ClickUp Ticket](https://app.clickup.com/t/29xd1x4)

### DIFF
--- a/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
+++ b/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
@@ -7,6 +7,7 @@ import com.codeforcommunity.dto.site.AdoptedSitesResponse;
 import com.codeforcommunity.dto.site.EditSiteRequest;
 import com.codeforcommunity.dto.site.NameSiteEntryRequest;
 import com.codeforcommunity.dto.site.ParentAdoptSiteRequest;
+import com.codeforcommunity.dto.site.ParentRecordStewardshipRequest;
 import com.codeforcommunity.dto.site.RecordStewardshipRequest;
 import com.codeforcommunity.dto.site.UpdateSiteRequest;
 import java.sql.Date;
@@ -24,7 +25,7 @@ public interface IProtectedSiteProcessor {
   /** Removes the record in the adopted sites table linking the site to its current adopter */
   void forceUnadoptSite(JWTData userData, int siteId);
 
-  /** Adopts a tree for the Child through the Parent Account **/
+  /** Adopts a tree for a child account through the parent account **/
   void parentAdoptSite(JWTData parentUserData, int siteId, ParentAdoptSiteRequest parentAdoptSiteRequest,
                        Date dateAdopted);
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/codeforcommunity/dto/site/RecordStewardshipRequest.java
```diff
@@ -24,7 +24,7 @@ public RecordStewardshipRequest(
     this.weeded = weeded;
   }
 
-  private RecordStewardshipRequest() {}
+  protected RecordStewardshipRequest() {}
```
priime0, line 27: nit: Maybe `default` would be better here instead?

api/src/main/java/com/codeforcommunity/dto/site/ParentRecordStewardshipRequest.java
```diff
@@ -0,0 +1,47 @@
...
+import java.sql.Date;
+import java.util.List;
+
+public class ParentRecordStewardshipRequest extends RecordStewardshipRequest {
+
+  private Integer childUserId;
+
+  public ParentRecordStewardshipRequest(
+      Date date,
+      boolean watered,
```
priime0, line 14: discussion: Maybe in the future we should put these activities (`watered`, `mulched`, `cleaned`, `weeded`) in an object. Pinging @chromium-52 . Might take significant refactoring though.

priime0 (commented):
lgtm :+1:

priime0 (approved):

api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
```diff
@@ -23,13 +25,22 @@ public interface IProtectedSiteProcessor {
   /** Removes the record in the adopted sites table linking the site to its current adopter */
   void forceUnadoptSite(JWTData userData, int siteId);
 
+  /** Adopts a tree for a child account through the parent account **/
```
rymaju, line 28: Are these purpose statements doing more than what the function name already provides?

api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
```diff
@@ -23,13 +25,22 @@ public interface IProtectedSiteProcessor {
...
+                       Date dateAdopted);
+
   /** Get users adopted sites */
   AdoptedSitesResponse getAdoptedSites(JWTData userData);
 
   /** Records a new stewardship activity in the stewardship table linked to the given site */
   void recordStewardship(
       JWTData userData, int siteId, RecordStewardshipRequest recordStewardshipRequest);
 
+  /** Records a new stewardship activity for a child account through a parent account */
```
rymaju, line 39: Same as above

api/src/main/java/com/codeforcommunity/dto/site/ParentRecordStewardshipRequest.java
rymaju: +1
iirc we do this to mirror how we eventually store this in the DB. Remember our persistence model need not be our program model.

api/src/main/java/com/codeforcommunity/dto/site/RecordStewardshipRequest.java
rymaju: We should throw more keywords at it until it works like good Java programmers. /s

rymaju (approved):
I didnt realize our website was so popular with children! Does SFTT get used in schools or something? 🤔

api/src/main/java/com/codeforcommunity/dto/site/ParentRecordStewardshipRequest.java
priime0: That does make sense.

PR #118, opened 2022-11-12
feat: add plantingDate to getSite route

priime0:
## Why

[ClickUp Ticket](https://app.clickup.com/t/2u0bffn)

Allows admins to specify the date a tree was planted. This affects the icon used when displaying the tree on the map.

## This PR

- Add plantingDate to AddSiteRequest
- Update response to include plantingDate

## Verification Steps

Planted a tree with a plantingDate 3 years + 1 day ago and a tree with a plantingDate 3 years - 1 day ago. Trees show the correct icon. The database reflects the correct date assigned.

Signed-off-by: Lucas Sta Maria <lucas@priime.dev>


api/src/main/java/com/codeforcommunity/dto/site/AddSiteRequest.java
```diff
@@ -12,6 +12,7 @@ public class AddSiteRequest extends UpdateSiteRequest {
   private String zip;
   private String address;
   private Integer neighborhoodId;
+  private java.sql.Date plantingDate;
```
chromium-52, line 15: :wrench: Since `plantingDate` is specific to a site entry (in other words, a site can have multiple site entries, each with a different planting date), could we move the `plantingDate` field to `UpdateSiteRequest`? Also, since we don't import anything else named `Date` here, we should just be able to do `import java.sql.Date;` at the top and do `private Date plantingDate`

api/src/main/java/com/codeforcommunity/dto/site/GetSiteResponse.java
```diff
@@ -13,6 +13,7 @@ public class GetSiteResponse {
   private final String address;
   private final Integer neighborhoodId;
   private final List<SiteEntry> entries;
+  private final java.sql.Date plantingDate;
```
chromium-52, line 16: :wrench: Similar to the comment above, can we move `plantingDate` to `SiteEntry.java` (src/main/java/com/codeforcommunity/dto/site/SiteEntry.java) so that a site can have multiple planting dates?

chromium-52 (commented):
:100: Looks good! Just one fix to move the planting dates from sites to site entries

api/src/main/java/com/codeforcommunity/dto/site/AddSiteRequest.java
rymaju: Agreed

api/src/main/java/com/codeforcommunity/dto/site/AddSiteRequest.java
```diff
@@ -167,6 +170,14 @@ public void setNeighborhoodId(Integer neighborhoodId) {
     this.neighborhoodId = neighborhoodId;
   }
 
+  public java.sql.Date getPlantingDate() {
```
rymaju, line 173:
```suggestion
  public Date getPlantingDate() {
```

api/src/main/java/com/codeforcommunity/dto/site/AddSiteRequest.java
```diff
@@ -167,6 +170,14 @@ public void setNeighborhoodId(Integer neighborhoodId) {
...
+    return this.plantingDate;
+  }
+
+  public void setPlantingDate(java.sql.Date plantingDate) {
```
rymaju, line 177:
```suggestion
  public void setPlantingDate(Date plantingDate) {
```

api/src/main/java/com/codeforcommunity/dto/site/GetSiteResponse.java
```diff
@@ -23,7 +24,8 @@ public GetSiteResponse(
       String zip,
       String address,
       Integer neighborhoodId,
-      List<SiteEntry> entries) {
+      List<SiteEntry> entries,
+      java.sql.Date plantingDate) {
```
rymaju, line 28:
```suggestion
      Date plantingDate) {
```

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -128,10 +128,18 @@ private List<SiteEntry> getSiteEntries(int siteId) {
   @Override
   public GetSiteResponse getSite(int siteId) {
     SitesRecord sitesRecord = db.selectFrom(SITES).where(SITES.ID.eq(siteId)).fetchOne();
+    SiteEntriesRecord siteEntriesRecord =
```
rymaju, line 131: 🔧 Move this query below the check on 137. The if(site_XXX == null) is essentially a post-condition check for the query on 130. You break the cohesion between these two lines by shoving this query between them.

🏕️ Honestly, both of these should be put into their own methods.

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -270,6 +270,7 @@ public void addSite(JWTData userData, AddSiteRequest addSiteRequest) {
     siteEntriesRecord.setStump(addSiteRequest.isStump());
     siteEntriesRecord.setTreeNotes(addSiteRequest.getTreeNotes());
     siteEntriesRecord.setSiteNotes(addSiteRequest.getSiteNotes());
+    siteEntriesRecord.setPlantingDate(addSiteRequest.getPlantingDate());
```
rymaju, line 273: 💭 I hate that we're still making people do this. Sadly, this is just the way the codebase is set up.

api/src/main/java/com/codeforcommunity/dto/site/GetSiteResponse.java
```diff
@@ -33,6 +35,7 @@ public GetSiteResponse(
     this.address = address;
     this.entries = entries;
     this.neighborhoodId = neighborhoodId;
+    this.plantingDate = plantingDate;
```
rymaju, line 38: 🔧 You're writing exactly the same code twice. We purposefully made these Plain Old Java Objects so that we can abstract freely. Take common code and extract them into an abstract class ala FII.

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -128,10 +128,18 @@ private List<SiteEntry> getSiteEntries(int siteId) {
...
+      db.selectFrom(SITE_ENTRIES)
+        .where(SITE_ENTRIES.SITE_ID.eq(siteId))
+        .orderBy(SITE_ENTRIES.UPDATED_AT.desc())
```
rymaju, line 134: ❓ Why order? Specify this in the code as a purpose statement.

rymaju (changes requested):

PR #119, opened 2022-11-13
Add Edit Stewardship Route

huang0h:
# Why

[ClickUp ticket](https://app.clickup.com/t/3ucv22f)

Add route to allow admins and tree site owners to edit existing stewardship activities

Route: ```POST api/v1/protected/sites/edit_stewardship/:site_id```

# This PR

- Add new class EditStewardShipRequest 
- Update ProtectedSiteRouter, ProtectedSiteProcessorImpl to handle edit stewardship requests

# Verification steps

- Request to edit existing record as admin - database correctly reflects changes
- Request to edit existing record as non-admin owner - database correctly reflects changes
- Request to edit nonexistent record - returns proper error status + message
- Request to edit existing record as non-admin/non-owner - returns proper error status + message
- Request to edit existing record with malformed body - returns proper error status + message

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
+++ b/api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
@@ -1,0 +1,43 @@
+package com.codeforcommunity.dto.site;
+
+import com.codeforcommunity.exceptions.HandledException;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EditStewardshipRequest extends RecordStewardshipRequest {
+
+  private Integer activityId;
+
+  public EditStewardshipRequest(
+      java.sql.Date date, boolean watered, boolean mulched, boolean cleaned, boolean weeded, Integer activityId) {
+    super(date, watered, mulched, cleaned, weeded);
+    this.activityId = activityId;
+  }
+
+  public Integer getActivityId() {
+    return activityId;
+  }
+
+  public void setActivityId(Integer activityId) {
+    this.activityId = activityId;
+  }
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "edit_stewardship_request.";
+    List<String> fields = new ArrayList<>();
+
+    if (getDate() == null) {
+      fields.add(fieldName + "date");
+    }
+    if (!(getWatered() || getMulched() || getCleaned() || getWeeded())) {
+      fields.add(fieldName + "activities");
+    }
+    if (activityId == null) {
+      fields.add(fieldName + "activityId");
+    }
+
+    return fields;
+  }
+}
--- a/api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
+++ b/api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
@@ -7,20 +7,9 @@ import java.util.List;
 
 public class EditStewardshipRequest extends RecordStewardshipRequest {
 
-  private Integer activityId;
-
   public EditStewardshipRequest(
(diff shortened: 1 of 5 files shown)


api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  private EditStewardshipRequest() {}
+
+  @Override
+  public List<String> validateFields(String fieldPrefix) throws HandledException {
+    String fieldName = fieldPrefix + "edit_stewardship_request.";
+    List<String> fields = new ArrayList<>();
+
+    if (getDate() == null) {
```
chromium-52, line 21: 🔧 Can we access the `date` field directly instead of using a getter so that it's consistent with the rest of the method (and other `validateFields` methods)? Also, since the field validation logic is the same for `RecordStewardshipRequest` and `EditStewardshipRequest`, we can abstract it out into a helper method.

api/src/main/java/com/codeforcommunity/dto/site/EditStewardshipRequest.java
```diff
@@ -0,0 +1,30 @@
+package com.codeforcommunity.dto.site;
+
+import com.codeforcommunity.exceptions.HandledException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class EditStewardshipRequest extends RecordStewardshipRequest {
+
+  public EditStewardshipRequest(
+      java.sql.Date date, boolean watered, boolean mulched, boolean cleaned, boolean weeded) {
```
chromium-52, line 10: 🔧 Since we don't import anything else named `Date` in this file, we should just be able to do `import java.sql.Date;` at the top and do `Date date` here

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -351,13 +363,33 @@ public void deleteSite(JWTData userData, int siteId) {
...
         db.selectFrom(STEWARDSHIP).where(STEWARDSHIP.ID.eq(activityId)).fetchOne();
 
-    if (activity == null) {
-      throw new ResourceDoesNotExistException(activityId, "Stewardship Activity");
+    if (!(activity.getUserId().equals(userData.getUserId())
+        || userData.getPrivilegeLevel().equals(PrivilegeLevel.SUPER_ADMIN)
+        || userData.getPrivilegeLevel().equals(PrivilegeLevel.ADMIN))) {
+      throw new AuthException(
+          "User needs to be an admin or the activity's author to edit the record.");
     }
```
chromium-52, line 377: 🔧 Since this logic is used twice, can we abstract this out? Also, the `isAdmin(PrivilegeLevel)` method (on line 107) can be used to simplify the logic that checks if the user is an admin or super admin.

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -75,6 +76,17 @@ private void checkNeighborhoodExists(int neighborhoodId) {
...
   }
 
+  /**
+   * Check if a stewardship record exists.
+   *
+   * @param activityId to check
+   */
+  private void checkStewardshipExists(int activityId) {
```
chromium-52, line 84: 💯

chromium-52 (commented):
👍 LGTM! Just a few minor suggestions

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
Tvcz: ^^

Tvcz (approved):
lgtm

chromium-52 (approved):
💯 LGTM!

PR #120, opened 2022-12-14
Add addSites route

chromium-52:
Dependent PR: https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/172

Modify the addSites route that allows admin users to add multiple new sites via a CSV file upload. The request body of the addSites request is a String representing the contents of the uploaded CSV file instead of a DTO (`AddSitesRequest.java`).

Add a new dependency called Jackson for parsing the contents of a CSV file into a list of `AddSiteRequest`s in the backend.

Test that calling the route as a non-admin user returns an error with a 401 (unauthorized) status code


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -436,15 +441,36 @@ public void editSite(JWTData userData, int siteId, EditSiteRequest editSiteReque
...
+  private List<AddSiteRequest> parseCSVString(String sitesCSV) throws HandledException {
+    try {
+      CsvMapper mapper = new CsvMapper();
+      CsvSchema schema = CsvSchema.emptySchema().withHeader();
+      MappingIterator<AddSiteRequest> sitesIterator =
+          mapper.readerFor(AddSiteRequest.class).with(schema).readValues(sitesCSV);
+      List<AddSiteRequest> addSiteRequests = sitesIterator.readAll();
+      for (AddSiteRequest addSiteRequest : addSiteRequests) {
+        addSiteRequest.validate();
+      }
```
huang0h, line 469: ⛏️ Could we use the same `List.forEach` method from line 449 to make this block a little more readable? Maybe something like:
```suggestion
      addSiteRequests.forEach(siteRequest -> siteRequest.validate());
```

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -436,15 +441,36 @@ public void editSite(JWTData userData, int siteId, EditSiteRequest editSiteReque
...
+      return addSiteRequests;
+    } catch (HandledException | IOException e) {
```
huang0h, line 471: ❓ Is there a specific reason why we catch an IOException to throw an InvalidCSVException? I feel like if something happens with I/O here, it may not always be due to a bad CSV file, unless this is how Jackson signals such a case.

huang0h (commented):
Looks good! 💯

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
Tvcz: Also curious about this ^^

Tvcz (approved):
lgtm

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
chromium-52: Yeah there are tons of reasons why an IOException could be thrown in general (note that the documentation for the two methods we have here that could throw an IOException, [readValues(String)](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectReader.html#readValues(java.lang.String)) and [readAll()](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/MappingIterator.html#readAll()), don't actually document when it can be thrown so there's no real way to know exactly when it can be thrown without digging through their code). The reason why I did this in this case was because we're reading from a string (rather than something like a file or an input stream) so that there's only so many reasons why an IOExeption could be thrown, most (if not all) of which is probably due to an invalid CSV file.

huang0h (approved):
lgtm!

PR #121, opened 2022-12-19
Update get site and name site entry routes

chromium-52:
Modify the get site route so that it also returns the tree name for each site entry.

Modify the name site entry route so that passing in an empty string deletes its name (i.e. sets is to `null`).


Tvcz (approved):
Lgtm

PR #124, opened 2023-01-07
Make email equals check case insensitive

chromium-52:
ClickUp ticket: https://app.clickup.com/t/47u12vb

Change checking for email equality to always be case insensitive.

Tested that the following affected routes still work:

* Log in
* Sign up
* Request password reset
* Change email
* Change privilege level
* Create child account


huang0h (approved):
Lgtm 👍

PR #140, opened 2023-02-07
Add addSites route

chromium-52:
## Checklist

## Why


rymaju:
Note that String is a pretty weak replacement for the DTO we used to have.

Ideally we would move the CSV parsing into a DTO that consumed the string, and we pass that around instead.

PR #144, opened 2023-02-20
Nd.ah.delete image

huang0h:
## Checklist

- [x] 1. Change ClickUp ticket status to "In Review"
- [x] 2. After making the PR, add PR link to ClickUp ticket

## Why

[ClickUp Ticket](https://app.clickup.com/t/862hy9y69)

Add route that allows admins to delete images associated with a site entry

## This PR

- Implement new route `POST api/v1/protected/sites/delete_image/:image_id`

## Verification Steps

- Verified that as an admin, can delete any image
- Verified that as a non-admin, cannot delete any image and request returns a 401 status code
- Verified that request to nonexistent images return a 400 status code


api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
```diff
@@ -79,4 +79,7 @@ void editStewardship(
    * is NULL, any pre-existing site image for the specified site will be deleted
    */
   void uploadSiteImage(JWTData userData, int siteId, UploadSiteImageRequest uploadSiteImageRequest);
+
+  /** Removes the given image */
```
chromium-52, line 83: ⛏️ Could we change this to "Removes the site image with the given imageId"?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -552,4 +565,15 @@ public void uploadSiteImage(
 
     // site.store();
   }
+
+  @Override
+  public void deleteSiteImage(JWTData userData, int imageId) {
+    checkImageExists(imageId);
+    SiteImagesRecord imagesRecord = db.selectFrom(SITE_IMAGES).where(SITE_IMAGES.ID.eq(imageId)).fetchOne();
+    if (!isAdmin(userData.getPrivilegeLevel())) {
```
chromium-52, line 573: 🔧 Could we use the `assertAdminOrSuperAdmin` method and call it before `checkImageExists`?

https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/blob/2c48f6568629d1e57f1a7da8260b1cff38d19476/service/src/main/java/com/codeforcommunity/processor/AbstractProcessor.java#L12-L16

chromium-52 (commented):
💯 Looks great! Just two quick comments

chromium-52 (approved):
👍 lgtm!

PR #157, opened 2023-04-14
Add email template

CerberusLatrans:
## Checklist

## Why

[ClickUp Ticket](https://app.clickup.com/t/43594vv)

Allows admins to upload html email templates to S3 for future re-use.

## This PR

Created a new subrouter, processor, dto for emailer (HTML content is encoded as a string).
S3Requester class updated to support HTML upload.
Uploads to sftt-user-uploads bucket to the email_templates folder

## Verification Steps

When logged in as admin, sending POST request to:
http://localhost:8081/api/v1/protected/emailer/add_template/
with body 
{
  "name": "some_name",
  "template": "your html content"
}

results in that HTML file being uploaded to the email_templates folder, with the admin id as metadata.


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -40,7 +40,10 @@
 import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
+<<<<<<< HEAD
+=======
 
+>>>>>>> e717a55a3b232e1ac2d042a7711011a08445d8ff
```
rymaju, line 46: 🔧 Bad merge resolution

rymaju (commented):
Whats the status on this PR @CerberusLatrans ?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +181,76 @@ private static String validateBase64ImageAndUploadToS3(
...
+   *
+   * @param name the desired name of the new file in S3 (without a file extension).
+   * @param directoryName the desired directory of the file in S3 (without leading or trailing '/').
+   * @param adminID the desired ID of the user uploading the HTML to S3.
+   * @param htmlContent the string encoding of the HTML to upload.
+   * @return null if the encoding fails validation and HTML file URL if the upload was successful.
+   * @throws BadRequestHTMLException if the string to HTML decoding failed.
+   * @throws S3FailedUploadException if the upload to S3 failed.
+   */
+  public static String uploadHTML(String name, String directoryName, Integer adminID, String htmlContent) {
```
chromium-52, line 214: 🔥

chromium-52 (commented):
Great job on this! Just a couple quick changes

Also, can you uncomment lines 18 through 20 of `deploy.py` as part of this PR?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +181,76 @@ private static String validateBase64ImageAndUploadToS3(
...
     String title =
         eventTitle.replaceAll(
             "[!@#$%^&*()=+./\\\\|<>`~\\[\\]{}?]", ""); // Remove special characters
     return title.replace(" ", "_").toLowerCase()
-        + "_thumbnail"; // The desired name of the file in S3
+        + suffix; // The desired name of the file in S3
+  }
+
+  /**
+   * Validate the given string encoding of HTML and upload it to the user upload S3 bucket.
```
huang0h, line 204: 💭 Do we want to look into making sure the HTML is valid beyond being decodeable, e.g. making sure tags are properly closed, all tags are recognized, etc.? Not sure how much effort that would take and if it's worth doing /cc @chromium-52

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +181,76 @@ private static String validateBase64ImageAndUploadToS3(
...
+   *
+   * @param name the desired name of the new file in S3 (without a file extension).
+   * @param directoryName the desired directory of the file in S3 (without leading or trailing '/').
+   * @param adminID the desired ID of the user uploading the HTML to S3.
+   * @param htmlContent the string encoding of the HTML to upload.
+   * @return null if the encoding fails validation and HTML file URL if the upload was successful.
```
huang0h, line 210: ⛏️ Looks like we don't actually return `null` anywhere, so can change this line to say something like `@return the HTML file URL after a successful upload`?

huang0h (commented):
Great work with setting up the S3 stuff! Looks good to me, just a couple comments:

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
chromium-52: I think some form of validation would be nice like making sure the HTML is valid

In cases where we can't trust users' inputs (which, to be fair, is always so that even in our case, this is somewhat but not necessarily an overkill), we can use libraries like [jsoup](https://jsoup.org/) or [OWASP Java HTML Sanitizer](https://github.com/OWASP/java-html-sanitizer) for input validation

api/src/main/java/com/codeforcommunity/rest/subrouter/EmailerRouter.java
```diff
@@ -0,0 +1,54 @@
+package com.codeforcommunity.rest.subrouter;
+
+import com.codeforcommunity.api.IEmailerProcessor;
+import com.codeforcommunity.api.ILeaderboardProcessor;
```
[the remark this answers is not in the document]
CerberusLatrans: oops-- not sure why these are in there my bad. My local right now for some reason has them removed.

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +181,76 @@ private static String validateBase64ImageAndUploadToS3(
...
+    String fullFileName = getFileNameWithoutExtension(name, "_template") + ".html";
+    File tempFile;
+
+    try {
+      // Temporarily writes the string to HTML file on disk
+      tempFile = File.createTempFile(fullFileName, null, null);
+      FileOutputStream fos = new FileOutputStream(tempFile);
+      byte[] bytesArray = htmlContent.getBytes();
+      fos.write(bytesArray);
+      fos.flush();
```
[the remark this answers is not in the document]
CerberusLatrans: Thanks for catching that!

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: You are right-- I copied a bit too much from the upload image method. Thanks for catching that

CerberusLatrans:
I made changes to address all of your suggestions. It now uses JSoup to to ensure only valid HTML can be uploaded, and the `BadReqestHTMLException`  message details any HTML (up to 10) errors that caused it to be thrown.

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +187,89 @@ private static String validateBase64ImageAndUploadToS3(
...
+   * @throws S3FailedUploadException if the upload to S3 failed.
+   */
+  public static String uploadHTML(
+      String name, String directoryName, Integer adminID, String htmlContent) {
+    // Save HTML to temp file
+    String fullFileName = getFileNameWithoutExtension(name, "_template") + ".html";
+    File tempFile;
+
+    // try to create a clean parse
+    Parser parser = Parser.htmlParser().setTrackErrors(10);
```
chromium-52, line 226: 💯

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -178,21 +187,89 @@ private static String validateBase64ImageAndUploadToS3(
...
     String title =
         eventTitle.replaceAll(
             "[!@#$%^&*()=+./\\\\|<>`~\\[\\]{}?]", ""); // Remove special characters
-    return title.replace(" ", "_").toLowerCase()
-        + "_thumbnail"; // The desired name of the file in S3
+    return title.replace(" ", "_").toLowerCase() + suffix; // The desired name of the file in S3
+  }
+
+  /**
+   * Validate the given string encoding of HTML and upload it to the user upload S3 bucket.
```
chromium-52, line 209: ⛏️ Can we add a sentence here that says that a template will be overwritten if another template with the same name is uploaded?

chromium-52 (approved):
👍lgtm and great job with jsoup! One really quick comment but looks good otherwise

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: Done!

PR #159, opened 2023-04-18
Edit site entries

CerberusLatrans:
## Checklist

- [x] 1. Change ClickUp ticket status to "In Review"
- [x] 2. After making the PR, add PR link to ClickUp ticket

## Why

[ClickUp Ticket](https://app.clickup.com/t/3ucuvx8)

Allows admins to edit site existing entries by specifying an entry Id and the updated values

## This PR

Uses the existing UpdateSiteRequest (name misleading) as dto, some filter site merge conflicts resolved.

## Verification Steps

Made a POST request in Postman: http://localhost:8081/api/v1/protected/sites/edit_entry/1
First with invalid access key to ensure admin only access.
Made a request with valid access key to update all values to null-- all values updated correctly, including updating userId and updatedAt

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
@@ -40,7 +40,6 @@ import com.codeforcommunity.exceptions.WrongAdoptionStatusException;
 import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
-
 import java.io.IOException;
 import java.sql.Date;
 import java.sql.Timestamp;
@@ -48,7 +47,6 @@ import java.time.LocalDate;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.stream.Collectors;
-
 import org.jooq.Condition;
 import org.jooq.DSLContext;
 import org.jooq.Result;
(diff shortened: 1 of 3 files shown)


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -642,26 +660,84 @@ SITE_ENTRIES.COMMON_NAME, USERS.FIRST_NAME, USERS.LAST_NAME, USERS.EMAIL, coales
...
+  public void editSiteEntry(JWTData userData, int entryId, UpdateSiteRequest editSiteEntryRequest) {
+    checkEntryExists(entryId);
+    assertAdminOrSuperAdmin(userData.getPrivilegeLevel());
+
+    SiteEntriesRecord siteEntriesRecord =
+        db.selectFrom(SITE_ENTRIES).where(SITE_ENTRIES.ID.eq(entryId)).fetchOne();
+
+    siteEntriesRecord.setUserId(userData.getUserId());
+    siteEntriesRecord.setUpdatedAt(new Timestamp(System.currentTimeMillis()));
+    siteEntriesRecord.setTreePresent(editSiteEntryRequest.isTreePresent());
```
huang0h, line 701: ⛏️  This giant block of code where we set all of the site entry fields already exists in two other places - can we pull those chunks out into its own method (and call it `populateSiteEntry` or something)? That way, if we ever add a new column to this table, we'll just need to change the one place where we fill in site entry data.

huang0h (commented):
💯 Looks good to me, just one comment on filling out site entries - make sure to update the [API spec doc](https://github.com/Code-4-Community/c4c-dev-docs) to say this route is implemented after you merge!

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
CerberusLatrans: Thanks for the tips Avery! I made a PR for the docs and refactored the duplicate code into the helper.

huang0h (approved):
lgtm!

chromium-52 (approved):
👍 lgtm!!

PR #175, opened 2023-05-28
Ot.delete template

CerberusLatrans:
## Checklist

## Why

Allows admins to delete existing email templates in the email-templates folder of the sftt-user-uploads bucket

## This PR

First checks if the file exists with `doesObjectExist`; however instead of returning `false` if the file does not exist, it throws an exception (seems to come from trying to access metadata) for a 403 forbidden error. Tried to enable attribute permissions on the bucket but doesn't work. Might have to instantiate S3 client with credentials for it to work? Works fine (returns `true`) with the file does actually exist.

There might be a better workaround, but I catch the S3 exception and throw an `InvaliURLException` instead.

## Verification Steps

Calling a POST request `http://localhost:8081/api/v1/protected/emailer/delete_template/[some existing name].html`
removes that file from the S3 folder

### DIFF
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
@@ -25,6 +25,7 @@ public class ProtectedEmailerRouter implements IRouter {
     Router router = Router.router(vertx);
 
     registerAddTemplate(router);
+    registerDeleteTemplate(router);
 
     return router;
   }
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
@@ -60,6 +60,7 @@ public class ProtectedEmailerRouter implements IRouter {
     LoadTemplateResponse loadTemplateResponse = processor.loadTemplate(userData, templateName);
 
     end(ctx.response(), 200, JsonObject.mapFrom(loadTemplateResponse).toString());
+  }
 
   private void registerDeleteTemplate(Router router) {
     Route deleteTemplate = router.post("/delete_template/:template_name");
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
@@ -63,7 +63,7 @@ public class ProtectedEmailerRouter implements IRouter {
   }
 
   private void registerDeleteTemplate(Router router) {
-    Route deleteTemplate = router.post("/delete_template/:template_name");
+    Route deleteTemplate = router.delete("/delete_template/:template_name");
     deleteTemplate.handler(this::handleDeleteTemplate);
   }
 
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
@@ -43,4 +43,18 @@ public class ProtectedEmailerRouter implements IRouter {
 
     end(ctx.response(), 200);
   }
+
+  private void registerDeleteTemplate(Router router) {
+    Route deleteTemplate = router.post("/delete_template/:template_name");
+    deleteTemplate.handler(this::handleDeleteTemplate);
+  }
+
+  private void handleDeleteTemplate(RoutingContext ctx) {
+    JWTData userData = ctx.get("jwt_data");
+    String templateName = RestFunctions.getRequestParameterAsString(ctx.request(), "template_name");
+
+    processor.deleteTemplate(userData, templateName);
+
+    end(ctx.response(), 200);
+  }
 }
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedEmailerProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedEmailerProcessorImpl.java
@@ -57,6 +57,7 @@ public class ProtectedEmailerProcessorImpl extends AbstractProcessor
     );
 
     return loadTemplateResponse;
+  }
 
   @Override
   public void deleteTemplate(JWTData userData, String templateName) {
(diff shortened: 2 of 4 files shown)


service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +275,31 @@ public static String uploadHTML(
...
+    } catch(AmazonServiceException e) {
+      throw new InvalidURLException();
+    }
+
+    // Create the request to delete the HTML
+    DeleteObjectRequest awsRequest = new DeleteObjectRequest(externs.getBucketPublic(), HTMLPath);
+
+    externs.getS3Client().deleteObject(awsRequest);
+
+    return String.format("%s/%s/%s", externs.getBucketPublicUrl(), directoryName, name);
```
huang0h, line 303: ⛏️ Do we use the return value anywhere? If not, I think we can just remove it, and make the method void instead.

huang0h (commented):
Overall LGTM, nice work! Most of my comments from the [load template PR](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/176) also apply here, specifically about the expected HTML path and adding AWS credentials.

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -325,4 +327,28 @@ public static LoadTemplateResponse loadHTML(String name, String directoryName) {
...
+   * @param name the name of the HTML file in S3 to be deleted.
+   * @param directoryName the directory of the file in S3 (without leading or trailing '/').
+   * @throws InvalidURLException if the file does not exist.
+   * @throws SdkClientException if the deletion from S3 failed.
+   */
+  public static void deleteHTML(String name, String directoryName) {
+    String HTMLPath = directoryName + "/" + name;
+
+    // if the object does not exist, it throws a 403 for metadata?
+    try {
```
chromium-52, line 343: 🔧 Now that we're initializing `s3Client` with the access/secret keys, we might not need this try-catch

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -325,4 +327,28 @@ public static LoadTemplateResponse loadHTML(String name, String directoryName) {
...
+  /**
+   * Delete the existing HTML file with the given name from the user uploads S3 bucket.
+   *
+   * @param name the name of the HTML file in S3 to be deleted.
+   * @param directoryName the directory of the file in S3 (without leading or trailing '/').
+   * @throws InvalidURLException if the file does not exist.
+   * @throws SdkClientException if the deletion from S3 failed.
+   */
+  public static void deleteHTML(String name, String directoryName) {
+    String HTMLPath = directoryName + "/" + name;
```
chromium-52, line 340: ⛏️ Could we append `_template.html` at the end of `htmlPath` here so that the caller won't have to specify that (for instance, if someone wants to delete a template called `some_name_template.html`, then they would only have to pass in `some_name` as the name of the template to delete)? The fact that we append `_template.html` to the name of the template when we store it in S3 is an internal implementation decision and if possible, we want to hide things like this from the user so that they won't have to care about it to make their life easier.

chromium-52 (commented):
lgtm! Just a few small comments

service/src/main/java/com/codeforcommunity/processor/ProtectedEmailerProcessorImpl.java
```diff
@@ -58,4 +58,10 @@ public LoadTemplateResponse loadTemplate(JWTData userData, String templateName)
 
     return loadTemplateResponse;
   }
+
+  @Override
+  public void deleteTemplate(JWTData userData, String templateName) {
+    assertAdminOrSuperAdmin(userData.getPrivilegeLevel());
+    S3Requester.deleteHTML(templateName, "email_templates");
```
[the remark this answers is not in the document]
CerberusLatrans: Yep, I forgot to change that once load template was merged

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: Good point-- I forgot to add this for delete template like with load and add template.

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -325,4 +327,23 @@ public static LoadTemplateResponse loadHTML(String name, String directoryName) {
...
+   * @param name the name of the HTML file in S3 to be deleted.
+   * @param directoryName the directory of the file in S3 (without leading or trailing '/').
+   * @throws InvalidURLException if the file does not exist.
+   * @throws SdkClientException if the deletion from S3 failed.
+   */
+  public static void deleteHTML(String name, String directoryName) {
+    String HTMLPath = directoryName + "/" + name + "_template.html";
+
+    Boolean htmlExists = externs.getS3Client().doesObjectExist(externs.getBucketPublic(), HTMLPath);
+
```
huang0h, line 343: 🔧 Can we use the `pathExists()` method here as well? We should also throw an `InvalidURLException` if the file doesn't exist

huang0h (commented):
Good work! Just a couple comments:

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedEmailerRouter.java
```diff
@@ -60,4 +61,18 @@ private void handleLoadTemplate(RoutingContext ctx) {
 
     end(ctx.response(), 200, JsonObject.mapFrom(loadTemplateResponse).toString());
   }
+
+  private void registerDeleteTemplate(Router router) {
+    Route deleteTemplate = router.post("/delete_template/:template_name");
```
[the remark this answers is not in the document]
CerberusLatrans: Sure that makes sense; I was just copying the style of the other routes according to the API like delete site or delete user. Do you think it makes sense to refactor these too?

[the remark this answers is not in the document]
huang0h: I think it's definitely a good housekeeping thing to do, but not that important right now - if we change the backend routes, we'll need to make changes in the frontend as well.

huang0h (approved):
lgtm!

PR #176, opened 2023-05-28
load email tempalte first iter

CerberusLatrans:
## Checklist

## Why

Allows admins to load existing email templates (retrieve the name, html content, author) in the email-templates folder of the sftt-user-uploads bucket

## This PR

First checks if the file exists with doesObjectExist; however instead of returning false if the file does not exist, it throws an exception (seems to come from trying to access metadata) for a 403 forbidden error. Tried to enable attribute permissions on the bucket but doesn't work. Might have to instantiate S3 client with credentials for it to work? Works fine (returns true) with the file does actually exist.

There might be a better workaround, but I catch the S3 exception and throw an InvaliURLException instead.

## Verification Steps

Making a GET request `http://localhost:8081/api/v1/protected/emailer/load_template/[some existing name].html`
responds with a json
'''
{
    "name": "[directory]/[name].html",
    "template": "[html content]",
    "author": "[admin id]"
}
'''

### DIFF
--- a/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
+++ b/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
@@ -1,16 +1,21 @@
 package com.codeforcommunity.requester;
 
+import com.amazonaws.AmazonServiceException;
 import com.amazonaws.SdkClientException;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.amazonaws.services.s3.model.CannedAccessControlList;
+import com.amazonaws.services.s3.model.GetObjectRequest;
 import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.PutObjectRequest;
+import com.amazonaws.services.s3.model.S3Object;
 import com.codeforcommunity.aws.EncodedImage;
+import com.codeforcommunity.dto.emailer.LoadTemplateResponse;
 import com.codeforcommunity.exceptions.BadRequestHTMLException;
 import com.codeforcommunity.exceptions.BadRequestImageException;
 import com.codeforcommunity.exceptions.S3FailedUploadException;
+import com.codeforcommunity.exceptions.InvalidURLException;
 import com.codeforcommunity.propertiesLoader.PropertiesLoader;
 
 import java.io.File;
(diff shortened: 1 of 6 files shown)


service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +278,45 @@ public static String uploadHTML(
...
+    S3Object HTMLFile = externs.getS3Client().getObject(awsRequest);
+    StringBuilder content = new StringBuilder();
+    try {
+      int c = 0;
+      while ((c = HTMLFile.getObjectContent().read()) != -1) {
+        content.append((char) c);
+      }
+    } catch(IOException e) {
+      throw new BadRequestHTMLException("HTML file could not be decoded to string");
+    }
```
huang0h, line 314: 💯

huang0h (commented):
Great job! Just a couple comments on the implementation:

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +278,45 @@ public static String uploadHTML(
...
+   * @throws SdkClientException if the loading from S3 failed.
+   */
+  public static LoadTemplateResponse loadHTML(String name, String directoryName) {
+    String HTMLPath = directoryName + "/" + name;
+
+    // if the object does not exist, it throws a 403 for metadata?
+    try {
+      Boolean htmlExists = externs.getS3Client().doesObjectExist(externs.getBucketPublic(), HTMLPath);
+    } catch(AmazonServiceException e) {
+      throw new InvalidURLException();
```
huang0h, line 299: I looked into this issue a little and it seems you're right about needing credentials here - when I added credentials to the client builder, `doesObjectExist()` worked as expected.

On line 36 in this file, before we call `build()` on the client builder, can we add this method call: `.withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials(PropertiesLoader.loadProperty("aws_access_key"), PropertiesLoader.loadProperty("aws_secret_key"))))`
This should provide the client with the proper credentials to check if objects exist and get `doesObjectExist()` working properly.

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +278,45 @@ public static String uploadHTML(
...
+   *
+   * @param name the name of the HTML file in S3 to be loaded.
+   * @param directoryName the directory of the file in S3 (without leading or trailing '/').
+   * @return LoadTemplateResponse with the html file name, content, and author.
+   * @throws InvalidURLException if the file does not exist.
+   * @throws BadRequestHTMLException if the HTML file cannot be decoded to a string.
+   * @throws SdkClientException if the loading from S3 failed.
+   */
+  public static LoadTemplateResponse loadHTML(String name, String directoryName) {
+    String HTMLPath = directoryName + "/" + name;
```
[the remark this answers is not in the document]
CerberusLatrans: Good idea-- In the frontend I'm guess that we wouldn't include the "_template" extension for display right?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: Thanks for this-- it works properly now (in the helper method)

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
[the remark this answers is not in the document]
huang0h: Yep, I think it's a bit more user-friendly to omit that part.

huang0h (approved):
LGTM - great work!

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +282,48 @@ public static String uploadHTML(
 
     return String.format("%s/%s/%s", externs.getBucketPublicUrl(), directoryName, name);
   }
+
+  // helper to check whether the given path exists
+  public static boolean pathExists(String path) {
+    Boolean exists = externs.getS3Client().doesObjectExist(externs.getBucketPublic(), path);
```
chromium-52, line 288: :pick: Can we turn this method into one line? Maybe something like `return externs.getS3Client().doesObjectExist(externs.getBucketPublic(), path);`

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +282,48 @@ public static String uploadHTML(
...
+      while ((c = HTMLFile.getObjectContent().read()) != -1) {
+        content.append((char) c);
+      }
+    } catch(IOException e) {
+      throw new BadRequestHTMLException("HTML file could not be decoded to string");
+    }
+
+    String HTMLContent = content.toString();
+
+    String HTMLAuthor = HTMLFile.getObjectMetadata().getUserMetaDataOf("userID");
```
chromium-52, line 325: :pick: Can we make `HTMLAuthor` the uploader's first (or full) name so that it'll be more recognizable by the admin in the frontend? But if we don't need to show their name, we might not need the author in the response. Also, even more of a nitpick but we can rename `HTMLAuthor` and the other variables to `htmlAuthor`, etc.

chromium-52 (approved):
:+1: Two small nitpicks but looks great to me!

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: I did the db querying in the processor impl class to get the first and last name. It is a little hacky in that I am isntantiating the LoadTemplateResponse dto twice (once from the S3 requester with the id as the author, and once after querying with the full name as the author to finally return).

PR #177, opened 2023-05-28
Ot.edit template

CerberusLatrans:
## Checklist

## Why

Allows admins to edit (overwrite) an existing email template in the email-templates folder of the sftt-user-uploads bucket

## This PR

First checks if the file exists with doesObjectExist; however instead of returning false if the file does not exist, it throws an exception (seems to come from trying to access metadata) for a 403 forbidden error. Tried to enable attribute permissions on the bucket but doesn't work. Might have to instantiate S3 client with credentials for it to work? Works fine (returns true) with the file does actually exist.

There might be a better workaround, but I catch the S3 exception and throw an InvaliURLException instead.

Also-- may not be optimal, but uses the existing `AddTemplateRequest` dto but ignores the `"name"` (only cares about `"template"`) attribute/key since that is taken from the query string `:template_name` instead.

## Verification Steps

Making a POST request http://localhost:8081/api/v1/protected/emailer/edit_template/[some existing name].html
responds with 200 and the object in S3's content has changed

### DIFF
--- a/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
+++ b/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
@@ -299,8 +299,9 @@ public class S3Requester {
     } catch(AmazonServiceException e) {
       throw new InvalidURLException();
     }
+    String name_no_suffix = name.substring(0, (name.length()-"_template.html".length()));
 
-    String overwriteResponse = uploadHTML(name, directoryName, adminID, htmlContent);
+    String overwriteResponse = uploadHTML(name_no_suffix, directoryName, adminID, htmlContent);
     return overwriteResponse;
   }
 }
--- a/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
+++ b/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
@@ -1,5 +1,6 @@
 package com.codeforcommunity.requester;
 
+import com.amazonaws.AmazonServiceException;
 import com.amazonaws.SdkClientException;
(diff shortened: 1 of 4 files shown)


service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -273,4 +275,33 @@ public static String uploadHTML(
 
     return String.format("%s/%s/%s", externs.getBucketPublicUrl(), directoryName, name);
   }
+
+  /**
+   * Validate the given string encoding of HTML and overwrite an existing file in the user upload S3 bucket.
+   *
```
huang0h, line 281: 💭 If the add template route already exists and overwrites templates with the same name, do we still need this route? Not totally sure how the UI will work for creating/updating templates, but I feel like we could use the add template route for both functions, considering the only difference here is that we first check if the file exists.

CerberusLatrans: Good idea-- I was also thinking about this and agree that they might be redundant. Since load, delete, and edit all check for the existence of the file, I could abstract this out into a separate route to be called before each one (therefore eliminating the edit template) but not sure if this is the best design.

chromium-52: Hmm I agree this route is redundant and we can probably close this PR. Sorry about that!

If the load and delete routes both check for the file's existence, we can probably just make a private helper method that does this and have both routes call that method. It's great that you noticed this tho!

CerberusLatrans:
redundant (with add template) PR

PR #185, opened 2023-06-11
Reset tree name when tree is unadopted or force unadopted

chromium-52:
## Checklist

## Why

[GH Projects Ticket](https://github.com/orgs/Code-4-Community/projects/15/views/1?pane=issue&itemId=30450878)

A user shouldn't be able to adopt a tree, name it something, and unadopt it while keeping its name, especially since there isn't a dedicated tree name review process and users can name any tree they adopt to anything they want atm.

## This PR

In both the unadopt and force unadopt routes, in addition to unadopting the specified site, query its latest site entry and reset its name (i.e. change it back to `null`).

## Verification Steps

Manually verified that unadopting or force unadopting a tree sets the name of its latest site entry to `null`

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
@@ -236,9 +236,23 @@ public class ProtectedSiteProcessorImpl extends AbstractProcessor
       throw new WrongAdoptionStatusException(false);
     }
 
-    db.deleteFrom(ADOPTED_SITES)
-        .where(ADOPTED_SITES.USER_ID.eq(userData.getUserId()))
-        .and(ADOPTED_SITES.SITE_ID.eq(siteId))
+    db.transaction(
+        configuration -> {
+          db.deleteFrom(ADOPTED_SITES)
+              .where(ADOPTED_SITES.USER_ID.eq(userData.getUserId()))
+              .and(ADOPTED_SITES.SITE_ID.eq(siteId))
+              .execute();
+
+          this.resetTreeName(siteId);
+        });
+  }
+
+  private void resetTreeName(int siteId) {
+    int latestSiteEntryId = this.latestSiteEntry(siteId).getId();
+
+    db.update(SITE_ENTRIES)
+        .setNull(SITE_ENTRIES.TREE_NAME)
+        .where(SITE_ENTRIES.ID.eq(latestSiteEntryId))
         .execute();
   }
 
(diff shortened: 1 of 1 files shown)


service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -551,6 +566,14 @@ public void nameSiteEntry(
     siteEntry.store();
   }
 
+  private SiteEntriesRecord latestSiteEntry(int siteId) {
+    return db.selectFrom(SITE_ENTRIES)
+        .where(SITE_ENTRIES.SITE_ID.eq(siteId))
+        .orderBy(SITE_ENTRIES.CREATED_AT.desc())
+        .fetchInto(SiteEntriesRecord.class)
+        .get(0);
```
huang0h, line 574: ⛏️ Right now this method assumes there's at least one entry for the site, which might not always be the case - calling this method with a site ID with no entry will throw an `IndexOutOfBoundsException`. Can we add a check to make sure we fetched at least one site entry?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -236,9 +236,23 @@ public void unadoptSite(JWTData userData, int siteId) {
...
+              .where(ADOPTED_SITES.USER_ID.eq(userData.getUserId()))
+              .and(ADOPTED_SITES.SITE_ID.eq(siteId))
+              .execute();
+
+          this.resetTreeName(siteId);
+        });
+  }
+
+  private void resetTreeName(int siteId) {
+    int latestSiteEntryId = this.latestSiteEntry(siteId).getId();
```
huang0h, line 251: ⛏️ Connecting with my other comment - in case there is no entry for the given site, can we break out of the function early so we don't call `getId()` on potentially nothing?

huang0h (commented):
Great work! Just a couple comments:

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
chromium-52: It's p counterintuitive but all sites actually (or should at least) have at least one site entry, regardless of whether a tree has ever been planted there or not (at the very least, a site where a tree has never been planted should have a single "site entry" with `treePresent` of `false`, like [this](https://map.treeboston.org/tree/3311014) site). And if a site has zero site entries, then I think it's a data quality issue and we should just add an "empty" site entry for that site. What's funny is that as of now, you can actually adopt, record a stewardship activity for, name, etc. an open planting site lol. Maybe we should take care of this in a separate PR

huang0h: Ah gotcha, thanks for clearing that up - yeah I feel like we shouldn't allow adopting open sites, since there's nothing to actually adopt there yet

huang0h (approved):
lgtm!

CerberusLatrans:
Looks gucci! and yeah the adoption of treeless site entry issue makes sense to fix

CerberusLatrans (approved):

PR #186, opened 2023-06-17
Connect edit site entry route with frontend

chromium-52:
## Checklist

## Why

[GH Projects Ticket](https://github.com/orgs/Code-4-Community/projects/15/views/1?pane=issue&itemId=29884931)

Add a new column called `CREATED_AT` to the `SITE_ENTRIES` table.

Frontend PR [here](https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/252)

## This PR

Add `CREATED_AT` to `SITE_ENTRIES` so that we'll be able to keep track of both when a site entry was added and last edited by an admin (instead of just when it was added).


CerberusLatrans:
Looks good to me overall-- just had one question

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
```diff
@@ -182,20 +182,20 @@ public SiteGeoResponse getSiteGeoJson() {
...
 
     Table<
             Record2<
                 Integer, // Site Entry ID
-                Timestamp // Updated At
+                Timestamp // Created At
             >>
-        recentlyUpdated =
+        recentlyCreated =
             table(
```
huang0h, line 193: 💭 Is this how the code gets auto-formatted? This almost feels harder to read than if it were on one line 🤔

huang0h (approved):
lgtm!

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
chromium-52: Yeah `mvn clean install` doesn't touch this file. We could look into a formatter/linter for the backend too

chromium-52:
> Looks good to me overall-- just had one question
@CerberusLatrans what was your question?

CerberusLatrans:
> > Looks good to me overall-- just had one question
> > @CerberusLatrans what was your question?
Screenshot 2023-06-24 at 10 26 59 PM

I'll just paste it here

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -672,7 +674,7 @@ public List<FilterSitesResponse> filterSites(
                     ADOPTED_SITES.USER_ID,
                     ADOPTED_SITES.DATE_ADOPTED,
                     max(STEWARDSHIP.PERFORMED_ON).as(STEWARDSHIP.PERFORMED_ON),
-                    max(SITE_ENTRIES.UPDATED_AT).as(SITE_ENTRIES.UPDATED_AT),
```
chromium-52, line 675: > > > Looks good to me overall-- just had one question
> > > @CerberusLatrans what was your question?
> 
> Screenshot 2023-06-24 at 10 26 59 PM
> I'll just paste it here

Huh. The `max(SITE_ENTRIES.UPDATED_AT)` would make it so that if a site has multiple site entries that fall into the same group (which could be possible because a site can have multiple site entries and we're not grouping by site entry ID), then it would just get the latest date at which one of those site entries was created. And I think we did this since users can only adopt the current (i.e. most recent) site entry for a given site, which we determine by looking at when site entries were created. So instead of there being a single entry with multiple `CREATED_AT`s, it's multiple distinct entries, each with its own `CREATED_AT`, that we want to get the max of. Does that make sense? I'd be happy to explain more

But looking at this again, I don't think we need the `CREATED_AT` column at all because it's not being used anywhere. @huang0h can you double check this? If we don't use it, we should probably just take this column out

PS: In your screenshot, the orange "Pending" means that the comment hasn't been published yet and you just have to add the comment before you click "Submit review" in the screenshot below

image

CerberusLatrans: Your explanation makes sense! Thanks.
Also, sorry I didn't submit the review -- for some reason I thought that just meant that approval was pending because I still had comments. Ill make sure i click submit in the future lol

CerberusLatrans (approved):

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
huang0h: Yeah, I think we can do without it for now - unless there's a specific case where we'd want to know when the entry was created rather than when it was last updated, it's probably not needed. I will say it's a little more understandable to have both a `created_at` column and an `updated_at` column, but it ultimately doesn't feel that important considering we've been doing fine without it.

chromium-52: @CerberusLatrans dw, github actually allows you to submit just a single comment or multiple comments at the same time (as part of a review) and the way you submit them is different so it can get a little confusing lol

@huang0h ah sorry meant to say that selecting the `CREATED_AT` column in the query here (on line 675 of `ProtectedSiteProcessorImpl.java`) may not be necessary because we don't use it after the query (we def need both the `CREATED_AT` and `UPDATED_AT` columns in general, just not here I think). I'll try taking it out and we can add it back if there are any issues

huang0h: 🤦 sounds good!

PR #196, opened 2023-06-28
add tree presence checking

### LINKED ISSUE #187: Prevent adoption for open planting sites
**Why**
- As mentioned in the comments of [this PR](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/185), it's currently possible to adopt and record stewardship activities for open planting sites with no trees
- This shouldn't be allowed, as it doesn't make sense to take care of a tree that isn't there

**What to do**
- Add a check to `adoptSite()` in `ProtectedsiteProcessorImpl` to make sure the corresponding site entry has a tree (check if `TREE_PRESENT` is true)

huang0h:
## Why

Resolves #187 

## This PR

- Adopting sites now checks if the associated entry has a tree present, throws an exception if not

## Verification Steps

- Correctly returns a 400 and doesn't adopt the site when attempting to adopt a site with no tree present 
- Correctly returns a 200 and adopts the site when attempting to adopt a site with a tree present

### DIFF
--- a/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
+++ b/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
@@ -36,6 +36,7 @@ import com.codeforcommunity.exceptions.AuthException;
 import com.codeforcommunity.exceptions.HandledException;
 import com.codeforcommunity.exceptions.InvalidCSVException;
 import com.codeforcommunity.exceptions.LinkedResourceDoesNotExistException;
+import com.codeforcommunity.exceptions.NoTreePresentException;
 import com.codeforcommunity.exceptions.ResourceDoesNotExistException;
 import com.codeforcommunity.exceptions.WrongAdoptionStatusException;
 import com.fasterxml.jackson.databind.MappingIterator;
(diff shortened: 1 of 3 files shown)


chromium-52 (approved):
I think we should have some sort of corresponding frontend changes like disabling or hiding the adopt buttons and hiding the tree name section but this lgtm!

huang0h:
@chromium-52 That sounds like a good idea, thanks for pointing that out!

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -764,6 +770,13 @@ public void editSiteEntry(JWTData userData, int entryId, UpdateSiteRequest editS
     siteEntriesRecord.setUserId(userData.getUserId());
     populateSiteEntry(siteEntriesRecord, editSiteEntryRequest);
 
-    siteEntriesRecord.store();
+    int siteId = siteEntriesRecord.getSiteId();
+
+    db.transaction(configuration -> {
+      siteEntriesRecord.store();
+      if (!editSiteEntryRequest.isTreePresent() && isAlreadyAdopted(siteId)) {
```
chromium-52, line 777: Could we make sure that the entry that's being edited is the latest (i.e. most recent) site entry for a site here? If we're editing a site entry that's not the latest, we don't need to worry about unadopting (and we shouldn't force unadopt)

chromium-52 (approved):
👍 lgtm!

CerberusLatrans (approved):
solid!

PR #199, opened 2023-07-02
Ah.get tree images

### LINKED ISSUE #183: Add default image to tree
Blocked by #182

**Why** 
- Every tree will have a default photo if no users have uploaded pictures of it yet. The default picture will be the tree leaf. 

**What to do**
- set the default image for all trees based on the following:
> See the link below for example:
https://tree-map.nycgovparks.org/tree-map/tree/2137477

- Make default pictures for all unknown trees the SFTT silhouette graphic we use in the background of the page.

### LINKED ISSUE #198: Site Images: Retrieve site images when fetching site data
**Why**
- Users should see approved site images when visiting a site

**What to do**
- Edit the `getSite()` method in `SiteProcessorImpl` (along with relevant DTOs) to also retrieve approved images for its associated site entries
- Image content should be:
```json
[{
    "imageId": INT | NULL,
    "uploaderUsername": STRING | NULL
    "uploadedAt": DATE | NULL,
    "imageUrl": STRING
},
...]
```
imageId, uploaderUsername, and uploadedAt will be null if the image is the default image for the tree's species

huang0h:
## Why

Resolves #198 
Resolves #183 

Adds to the get site route to also retrieve user-uploaded images for the site

## This PR

- Adds the `SiteEntryImage` DTO to represent site image information
- Update `getSite()` to also retrieve site images
- Assigned default images to tree species
- Updated import router to import species' default images

## Verification Steps

- Verified a site with images will correctly fetch only approved user-uploaded images
- Verified a site without images, but whose tree has a default image, fetches said default image
- Verified a site without images nor default image returns an empty list
- Verified site images with anonymous user returns "Anonymous" instead of the uploader's username


huang0h:
As the default images for tree species live in the table `tree_species`, I updated the import router so you can import which species get which images. You can call the route `POST api/v1/protected/import/tree_species` with [this json file](https://s3.us-east-2.amazonaws.com/sftt.import.data/tree_species.json) as its body (you might need to clear the `tree_species` table first)

CerberusLatrans:
> As the default images for tree species live in the table `tree_species`, I updated the import router so you can import which species get which images. You can call the route `POST api/v1/protected/import/tree_species` with [this json file](https://s3.us-east-2.amazonaws.com/sftt.import.data/tree_species.json) as its body (you might need to clear the `tree_species` table first)

I don't think I have access to see the json:
Screenshot 2023-07-09 at 4 27 44 PM

persist/src/main/resources/db/migration/V4_8__AddTreeSpeciesDefaultImage.sql
CerberusLatrans, line 1: Is 500 characters ensured to be enough for any tree URL? Not sure if there is some AWS url length specification for this

CerberusLatrans: Also, if the default images are also being stored in AWS do you think it makes sense for them to have an image ID too?

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
CerberusLatrans, line 1: Looks good! Do we have any system in place for actually approving user-submitted images?

huang0h:
> > As the default images for tree species live in the table `tree_species`, I updated the import router so you can import which species get which images. You can call the route `POST api/v1/protected/import/tree_species` with [this json file](https://s3.us-east-2.amazonaws.com/sftt.import.data/tree_species.json) as its body (you might need to clear the `tree_species` table first)
> 
> I don't think I have access to see the json: Screenshot 2023-07-09 at 4 27 44 PM

Just sent the file to you on slack (and updated the link to take you to the object page instead) - sorry about that!

CerberusLatrans (approved):

service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -163,14 +172,55 @@ record -> {
                   record.getRemovalDate(),
                   record.getScientificName(),
                   record.getBiocharAdded(),
-                  record.getLastEditedUser());
+                  record.getLastEditedUser(),
+                  getSiteEntryImages(record.getId(), record.getCommonName()));
```
chromium-52, line 176: ⛏️ maybe move this to line 137 because it's not a Cambridge field

chromium-52 (approved):
👍 lgtm! Nice work!

Issue #181, opened 2023-06-04
Calculate tree benefits for a site entry


fzyang1227:
**Why**
- Show benefits for each tree planted based off species/size of the tree
- Allows user to see effect of planting such a tree

**What to do**
> Documentation: https://docs.c4cneu.com/sftt/sftt-api-spec/#calculate-tree-benefits

1. Create a new class called TreeBenefitsCalculator that calculates the environmental benefits of the given site entry (using its tree species and diameter)
2. Calculate, for each tree, energy conserved, stormwater filtered, air quality improved, and carbon dioxide removed
3. Reference and example calculation: https://github.com/OpenTreeMap/otm-ecoservice#example-calculation

For each returned value, the corresponding columns to add/aggregate are:
- energy: natural_gas , electricity 
- stormwater: hydro_interception 
- airQuality: all columns prefixed with aq 
- co2Removed: co2_sequestered, co2_avoided 
- co2Stored: co2_storage 

*Great feature for actually writing unit tests (but you don't have to)*

PR #200, opened 2023-07-02
Ot.calc tree benefits

CerberusLatrans:
## Why

Clickup link: https://app.clickup.com/t/862j16xrm
#181 

Automatically calculates and returns tree benefits for a given site

## This PR
Based on this calculation: https://github.com/OpenTreeMap/otm-ecoservice#example-calculation

Uses the 'common_name' instead of 'species' attribute to get the species code since I have noticed that in the Site Entries table, the species name is often null in the table when the common name is not.

Still, there are cases when both are null or the diameter is null so need to figure out how to handle this.

Also, maybe should round values before returning?

API [spec](https://docs.c4cneu.com/sftt/sftt-api-spec/#calculate-tree-benefits) says to give a site id, but would it make more sense to give a site entry id instead so that we can calculate the benefits of previous entries at sites?

## Verification Steps

I picked a random tree from opentreemap: https://www.opentreemap.org/boston/features/3298491/
It is a honeylocust of diameter 2.4 in.
Screenshot 2023-07-01 at 11 35 31 PM

I queried a for a honeylocust of diameter 2.4 in our database (44 results) and took the first one with entry id 2395. It has site id 3255878, so I called GET http://localhost:8081/api/v1/sites/3255878/calculate_benefits
and got this: 
Screenshot 2023-07-02 at 12 46 33 AM

The values are a little off, so currently debugging.
Incorrect values:
co2 stored/money (They have to date, but my calculations would be per year)
energy money


service/src/main/java/com/codeforcommunity/processor/SiteProcessorImpl.java
```diff
@@ -230,4 +232,33 @@ public List<String> getAllCommonNames() {
         .orderBy(SITE_ENTRIES.COMMON_NAME.asc())
         .fetchInto(String.class);
   }
+
+  @Override
+  public TreeBenefitsResponse calculateBenefits(int siteId) {
+    checkSiteExists(siteId);
+
+    SiteEntriesRecord record =
+        db.selectFrom(SITE_ENTRIES).where(SITE_ENTRIES.SITE_ID.eq(siteId)).fetchOne();
```
CerberusLatrans, line 241: Should probably make sure that the site entry is the latest entry for the site

CerberusLatrans: Fixed

service/src/main/java/com/codeforcommunity/requester/TreeBenefitsCalculator.java
```diff
@@ -0,0 +1,189 @@
...
+    double y0 =
+        db.selectFrom(TREE_BENEFITS)
+            .where(TREE_BENEFITS.SPECIES_CODE.eq(speciesCode))
+            .and(TREE_BENEFITS.DIAMETER.eq(x0))
+            .fetchOne(property, Double.class);
+    double y1 =
+        db.selectFrom(TREE_BENEFITS)
+            .where(TREE_BENEFITS.SPECIES_CODE.eq(speciesCode))
+            .and(TREE_BENEFITS.DIAMETER.eq(x1))
+            .fetchOne(property, Double.class);
```
huang0h, line 87: ⛏️ Instead of querying our database each time we call `calcProperty()`, could we instead fetch once in the constructor, store those records as properties, and fetch the given field from those records in this method? That way, we won't have to keep fetching the same information multiple times. The signature won't need to change, since we can call `get(Field property)` on a database record.

service/src/main/java/com/codeforcommunity/requester/TreeBenefitsCalculator.java
```diff
@@ -0,0 +1,189 @@
...
+
+public class TreeBenefitsCalculator {
+  private final DSLContext db;
+  final String speciesCode;
+  final double diameterCM;
+  static final double[] intervals =
+      new double[] {3.81, 11.43, 22.86, 38.10, 53.34, 68.58, 83.82, 99.06, 114.30};
+  final double x0;
+  final double x1;
+  static final Map<String, Double> currencyConversion =
```
huang0h, line 21: 💯

huang0h (commented):
Good work, thanks for taking this on!

service/src/main/java/com/codeforcommunity/requester/TreeBenefitsCalculator.java
```diff
@@ -0,0 +1,189 @@
...
+    this.x1 = x1;
+  }
+
+  // helper to query the species code from the tree_species table
+  private String getSpeciesCode(String commonName) {
+    TreeSpeciesRecord record =
+        db.selectFrom(TREE_SPECIES).where(TREE_SPECIES.COMMON_NAME.eq(commonName)).fetchOne();
+    if (record == null) {
+      throw new IllegalArgumentException("Unable to find entry with name " + commonName);
+    }
```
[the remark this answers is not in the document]
CerberusLatrans: good idea!

service/src/main/java/com/codeforcommunity/requester/TreeBenefitsCalculator.java
CerberusLatrans: you are right-- I refactored

service/src/main/java/com/codeforcommunity/requester/TreeBenefitsCalculator.java
```diff
@@ -0,0 +1,189 @@
...
+      if (intervals[i] <= diameterCM) {
+        x0 = intervals[i];
+        x1 = intervals[i + 1];
+      }
+    }
+    if (x0 == 0 || x1 == 0) {
+      throw new IllegalArgumentException("diameter must be non-zero");
+    }
+    this.x0 = x0;
+    this.x1 = x1;
```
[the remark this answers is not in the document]
CerberusLatrans: Thanks for pointing that out. I didn't actually see their implementation before. When we have to clamp, that means interpolation will break (division by 0 since x0=x1), so I will just return y0 (or y1) whenever x0==x1

huang0h (approved):
lgtm, great job!!

PR #212, opened 2023-10-22
Knot.get emailer templates

### LINKED ISSUE #197: Emailer: Add route to fetch email template names
**Why**
- Admins should be able to see what templates are available to choose to load in the emailer

**What to do**
- Add a route `GET api/v1/protected/emailer/templates` to retrieve the names of the templates in S3
    - see [AmazonS3.listObjectsV2()](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3.html#listObjectsV2-com.amazonaws.services.s3.model.ListObjectsV2Request-)
- Return the names as a list of strings

CerberusLatrans:
## Why

Resolves #197

Gets all of the template names (with .html extension) in a list

## This PR
Kept .html extension, doesn't go into prod folder in s3.
`/emailer/template_names` instead of `/emailer/templates` in accordance with API doc

## Verification Steps

postman requested `http://localhost:8081/api/v1/protected/emailer/template_names`
set upload dir to prod/cambridge, got the list of files in there correctly

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/imports/TreeSpeciesImport.java
+++ b/api/src/main/java/com/codeforcommunity/dto/imports/TreeSpeciesImport.java
@@ -13,7 +13,8 @@ public class TreeSpeciesImport extends ApiDto {
 
   private String defaultImage;
 
-  public TreeSpeciesImport(String genus, String species, String commonName, String speciesCode, String defaultImage) {
+  public TreeSpeciesImport(
+      String genus, String species, String commonName, String speciesCode, String defaultImage) {
     this.genus = genus;
     this.species = species;
     this.commonName = commonName;
--- a/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
+++ b/service/src/main/java/com/codeforcommunity/requester/S3Requester.java
@@ -378,11 +378,16 @@ public class S3Requester {
     externs.getS3Client().deleteObject(awsRequest);
   }
 
+  /**
+   *
+   * @param bucketName
+   * @return
+   */
   public static List<String> getAllNamesinBucket(String bucketName) {
     ListObjectsV2Request req = new ListObjectsV2Request();
     req.setBucketName(bucketName);
     req.setDelimiter("/");
-    String path = String.join("/", TEMPLATE_S3_DIR, externs.getDirPublic());
+    String path = String.join("/", TEMPLATE_S3_DIR, externs.getDirPublic()+"/");
     req.setPrefix(path);
     req.setStartAfter(path);
 
@@ -392,7 +397,7 @@ public class S3Requester {
       res = externs.getS3Client().listObjectsV2(req);
       String prefix = req.getPrefix();
       List<String> result;
-      result = res.getObjectSummaries().stream().map(s -> s.getKey().replace(prefix, "")).collect(Collectors.toList());
+      result = res.getObjectSummaries().stream().map(s -> s.getKey().replace(prefix, "").replace("_template.html", "")).collect(Collectors.toList());
       return result;
     } catch (SdkClientException a) {
       throw new InvalidURLException();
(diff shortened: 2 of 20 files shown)


api/src/main/java/com/codeforcommunity/dto/imports/TreeSpeciesImport.java
```diff
@@ -13,7 +13,8 @@ public class TreeSpeciesImport extends ApiDto {
 
   private String defaultImage;
 
-  public TreeSpeciesImport(String genus, String species, String commonName, String speciesCode, String defaultImage) {
+  public TreeSpeciesImport(
+      String genus, String species, String commonName, String speciesCode, String defaultImage) {
```
huang0h, line 17: ❓ Is there a way to limit the autoformatter to specific files? It's definitely nice to have our code look good, but a bunch of extra files end up in the PR, and it can be hard to understand what the actual changes made are vs. just format changes

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -369,4 +377,28 @@ public static void deleteHTML(String name) {
 
     externs.getS3Client().deleteObject(awsRequest);
   }
+
+  public static List<String> getAllNamesinBucket(String bucketName) {
+    ListObjectsV2Request req = new ListObjectsV2Request();
+    req.setBucketName(bucketName);
+    req.setDelimiter("/");
+    String path = String.join("/", TEMPLATE_S3_DIR, externs.getDirPublic());
```
huang0h, line 385: 🔧 When I tried this route, it only returned an empty list until I added "/" to the end of the path - were you able to get template names without it?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -369,4 +377,28 @@ public static void deleteHTML(String name) {
...
+    req.setStartAfter(path);
+
+    ListObjectsV2Result res;
+    try {
+
+      res = externs.getS3Client().listObjectsV2(req);
+      String prefix = req.getPrefix();
+      List<String> result;
+      result = res.getObjectSummaries().stream().map(s -> s.getKey().replace(prefix, "")).collect(Collectors.toList());
+      return result;
```
huang0h, line 396: ⛏️ Can we also remove the template extension from the object names as well? This way, the frontend won't have to worry about removing the extensions, and it'll also match how we add and load templates - right now, those take just the template name (e.g. `test_email` rather than `test_email_template.html`)

huang0h (commented):
Great work! Thanks for taking this on 🥳

api/src/main/java/com/codeforcommunity/dto/imports/TreeSpeciesImport.java
CerberusLatrans: do you think the solution would be to try to only add the files we meant to change before committing?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: fixed and tested

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: fixed and tested

api/src/main/java/com/codeforcommunity/dto/imports/TreeSpeciesImport.java
huang0h: > do you think the solution would be to try to only add the files we meant to change before committing?

Yeah that would work too - just make sure to keep note of what files you changed, so you don't forget to add all the files you worked on

huang0h (approved):
lgtm!

SurabhiKeesara (approved):
Looks good! This is nitpicky, but try to remove spacing changes (just to clean up the PR)

PR #217, opened 2023-11-19
Clot.email attchment

### LINKED ISSUE #201: Ability to send emails with attachments
**Why**
- Might need to send out newsletters/other attachments with emails as admins

**What to do**
- see: https://www.simplejavamail.org/features.html#section-attachments
- Add the above to the emailing server

CerberusLatrans:
## Why

Resolves #201 

This allows admins to send a lot of different types of attachments to volunteers.

## This PR
Converts base64 to bytearraydata (which takes the mime type and data as byte[]) to use simplejavamail api. 
Frontend devs need to convert attachment to base 64 before using route.

## Verification Steps
Sent emails with jpg, docx, csv, zip, pdf


api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+    private String data;
+
+    public EmailAttachment(String name, String data) {
+        this.name = name;
+        this.data = data;
+    }
+
+    private EmailAttachment() {}
+
+    private static String getFileExtension(String base64Image) {
```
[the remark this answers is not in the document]
huang0h: Adding onto that, since this method is getting a file's mime type, could we rename it so it better describe what it's doing - maybe `getFileMimeType`?

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+        String[] dataSplit = metaSplit[0].split(":", 2); // Split the data type
+        if (dataSplit.length != 2) {
+            return null;
+        }
+
+        //String[] data = dataSplit[1].split("/", 2); // Split the image type here (e.g. "image/png")
+        //if (data.length != 2 || !data[0].equals("image")) {
+        //    // Ensure the encoded data is an image
+        //    return null;
+        //}
```
huang0h, line 58: ⛏️ Remember to clean up any blocks that are commented out! If you don't need this code, make sure to delete it

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+        String fileExtension = dataSplit[1]; // The image type (e.g. "png")
+        return fileExtension;
+    }
+    private static ByteArrayDataSource stringToDataSource(String data) {
+        byte[] imageData;
+        String mimeType;
+
+        try {
+            // Temporarily writes the image to disk to decode
+            mimeType = getFileExtension(data);
```
huang0h, line 69: ⛏️ It looks like `getFileExtension()` could potentially return `null`, but we don't ever check for that here. Once we call `getFileExtension()`, we should check if its result is null (in which case we should probably throw an exception) so we deal with the failure as soon as possible, rather than have it break something else down the line.

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+            System.out.println(mimeType);
```
huang0h, line 70: ⛏️ Similar to the commented-out blocks of code - make sure to remove any prints before you merge! If you think this information is important enough to print out, we can use the `SLogger.java` class for more robust logging

service/src/main/java/com/codeforcommunity/processor/ProtectedNeighborhoodsProcessorImpl.java
```diff
@@ -33,8 +37,10 @@ public void sendEmail(JWTData userData, SendEmailRequest sendEmailRequest) {
 
     String emailSubject = sendEmailRequest.getEmailSubject();
     String emailBody = sendEmailRequest.getEmailBody();
+    List<AttachmentResource> attachments = sendEmailRequest.getAttachments().stream()
+            .map(ea -> ea.getAttachmentResource()).collect(Collectors.toList());
```
huang0h, line 41: 💯

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -66,8 +58,10 @@ private static ByteArrayDataSource stringToDataSource(String data) {
 
         try {
             // Temporarily writes the image to disk to decode
-            mimeType = getFileExtension(data);
-            System.out.println(mimeType);
+            mimeType = getFileMimeType(data);
+            if (mimeType==null) {
+                throw new MalformedParameterException("data");
```
huang0h, line 63: Didn't know we had this exception, good find!

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,12 @@
...
+  "modulesFolders": [
+    "node_modules"
+  ],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
huang0h, line 12: 🔧 If you'd like, you can remove this folder along with `package-lock.json` and `yarn.lock` - not sure how these ended up here but they shouldn't affect anything.

huang0h (approved):
lgtm!

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public EmailAttachment(String name, String data) {
+        this.name = name;
+        this.data = data;
+    }
+
+    private EmailAttachment() {}
+
+    private static String getFileMimeType(String base64File) {
+        // Expected Base64 format: "data:image/{extension};base64,{imageData}"
```
SurabhiKeesara, line 28: Same comment as on reject email - avoid using comments and instead focus on good names (which it looks like you have, so you can delete there)

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,105 @@
+package com.codeforcommunity.dto.emailer;
```
SurabhiKeesara, line 1: All the same comments as on reject email
-  avoid using comments and instead focus on good names (which it looks like you have, so you can delete these)
- baseEncoding validation based on length

service/src/main/java/com/codeforcommunity/requester/Emailer.java
```diff
@@ -121,14 +121,18 @@ public void sendInviteTeamEmail(String sendToEmail, String sendToName, String te
...
-  public void sendArbitraryEmail(HashSet<String> sendToEmails, String subject, String body) {
+  public void sendArbitraryEmail(HashSet<String> sendToEmails, String subject, String body,
+                                 List<AttachmentResource> attachments) {
     String filePath = "/emails/Email.html";
 
     Map<String, String> templateValues = new HashMap<>();
     templateValues.put("body", body.replaceAll("\n", "<br />"));
     Optional<String> emailBody = emailOperations.getTemplateString(filePath, templateValues);
 
+
```
SurabhiKeesara, line 132: Super nitpicky but avoid adding extraneous spaces

SurabhiKeesara (commented):
Looks great, just a couple of comments

CerberusLatrans:
Retested both emailing with attachments and rejecting images after merge

PR #222, opened 2024-02-04
Clot.reject image

### LINKED ISSUE #219: Add Reject Site Image Route
**Why**
- In addition to deleting the image for rejection, we want to save a reason for rejection
- Allows admins to give users feedback on their images uploaded through an email, maybe inappropriate or could be a better image 

**What to do**
- Implement the route following the [API ref](https://docs.c4cneu.com/sftt/sftt-api-spec/#reject-site-image-admin-only)
- Create a new email template in `common/src/resources/emails` to fill with a reason, if one is provided
  - Look at `EmailChangeConfirmation.html` as an example - the title and body will need to be changed
- utilize the delete site image route logic to delete the given image

CiciLing:
## Why

Resolves #219 

This allows the admins to send rejection email to the users who uploaded inappropriate images. 

## This PR
We added a new email template "RejectImageEmail.html" to be used for image rejection email. 

## Verification Steps

Note that it will only send email when a rejection reason is given. When a reason is not given, it will delete the user image in database by using deleteSiteImage without sending the email.

### DIFF
--- a/api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
+++ b/api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
@@ -27,37 +27,33 @@ public class EmailAttachment extends ApiDto {
     private EmailAttachment() {}
 
     private static String getFileExtension(String base64Image) {
-        // Expected Base64 format: "data:image/{extension};base64,{imageData}"
-
         if (base64Image == null || base64Image.length() < 10) {
             return null;
         }
 
-        String[] base64ImageSplit = base64Image.split(",", 2); // Split the metadata from the image data
+        String[] base64ImageSplit = base64Image.split(",", 2);
         if (base64ImageSplit.length != 2) {
             return null;
         }
 
-        String meta = base64ImageSplit[0]; // The image metadata (e.g. "data:image/png;base64")
-        String[] metaSplit = meta.split(";", 2); // Split the metadata into data type and encoding type
+        String meta = base64ImageSplit[0];
+        String[] metaSplit = meta.split(";", 2);
 
         if (metaSplit.length != 2 || !metaSplit[1].equals("base64")) {
-            // Ensure the encoding type is base64
             return null;
         }
 
-        String[] dataSplit = metaSplit[0].split(":", 2); // Split the data type
+        String[] dataSplit = metaSplit[0].split(":", 2);
         if (dataSplit.length != 2) {
             return null;
         }
 
-        String[] data = dataSplit[1].split("/", 2); // Split the image type here (e.g. "image/png")
+        String[] data = dataSplit[1].split("/", 2);
         if (data.length != 2 || !data[0].equals("image")) {
-            // Ensure the encoded data is an image
             return null;
         }
 
-        String fileExtension = data[1]; // The image type (e.g. "png")
+        String fileExtension = data[1];
         return fileExtension;
     }
     private static ByteArrayDataSource stringToDataSource(String data) {
@@ -65,9 +61,7 @@ public class EmailAttachment extends ApiDto {
         String mimeType;
 
         try {
-            // Temporarily writes the image to disk to decode
             mimeType = getFileExtension(data);
-            System.out.println(mimeType);
             String[] base64Split = data.split(",", 2);
             imageData = Base64.getDecoder().decode(base64Split[1]);
 
--- a/api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
+++ b/api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
@@ -1,22 +1,69 @@
 package com.codeforcommunity.dto.emailer;
 
-import javax.mail.util.ByteArrayDataSource;
+import com.codeforcommunity.aws.EncodedImage;
+import com.codeforcommunity.exceptions.BadRequestImageException;
+import org.simplejavamail.api.email.AttachmentResource;
 
-public class EmailAttachment {
-    private final String name;
-    private final ByteArrayDataSource data;
+import javax.mail.util.ByteArrayDataSource;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Base64;
+
+public class EmailAttachment extends AttachmentResource {
 
 
     public EmailAttachment(String name, ByteArrayDataSource data) {
-        this.name = name;
-        this.data = data;
-
+        super(name, data);
     }
 
-    public String getName() {
-        return name;
+    private static String getFileExtension(String base64Image) {
+        // Expected Base64 format: "data:image/{extension};base64,{imageData}"
+
+        if (base64Image == null || base64Image.length() < 10) {
+            return null;
+        }
+
+        String[] base64ImageSplit = base64Image.split(",", 2); // Split the metadata from the image data
+        if (base64ImageSplit.length != 2) {
+            return null;
+        }
+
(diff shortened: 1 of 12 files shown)


api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public EmailAttachment(String name, String data) {
+        this.name = name;
+        this.data = data;
+    }
+
+    private EmailAttachment() {}
+
+    private static String getFileExtension(String base64Image) {
+        // Expected Base64 format: "data:image/{extension};base64,{imageData}"
```
SurabhiKeesara, line 30: Just cleanup comments throughout, the code is generally understandable when there are good naming conventions

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+
+        if (base64Image == null || base64Image.length() < 10) {
```
SurabhiKeesara, line 32: If the encoding length is 11 will it still be valid? If this is based on legacy code that's fine, just wondering if there's other ways to check image validity

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
```diff
@@ -0,0 +1,111 @@
...
+        return fileExtension;
+    }
+    private static ByteArrayDataSource stringToDataSource(String data) {
+        byte[] imageData;
+        String mimeType;
+
+        try {
+            // Temporarily writes the image to disk to decode
+            mimeType = getFileExtension(data);
+            System.out.println(mimeType);
```
SurabhiKeesara, line 70: Do we want to print this out?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -915,4 +922,24 @@ public void approveSiteImage(JWTData userData, int imageID) {
...
+                      .from(SITE_IMAGES)
+                      .where(SITE_IMAGES.ID.eq(imageId))
+                      .fetchOne(0, int.class);
+
+      UsersRecord user = db.selectFrom(USERS).where(USERS.ID.eq(uploaderId)).fetchOne();
+      String userEmail = user.getEmail();
+      String userFullName =
+              AuthDatabaseOperations.getFullName(user.into(Users.class));
+      emailer.sendRejectImageEmail(userEmail, userFullName, rejectImageRequest.getRejectionReason());
+    }
```
huang0h, line 942: ⛏️ I think we should send a rejection email regardless of if a reason was added just so to keep users updated - we could probably just add a default rejection message. Also, can we check if the reason is an empty string as well so we don't accidentally send no message?
Sorry if the API spec wasn't clear about this - that's totally my bad

common/src/main/resources/emails/RejectImageEmail.html
```diff
@@ -0,0 +1,392 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional //EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
SurabhiKeesara, line 1: This is a bit long. It's good for now, since we want to send this out. Just a follow up on our conversation to create a ticket for making email customization easier down the line

common/src/main/resources/emails/RejectImageEmail.html
```diff
@@ -0,0 +1,392 @@
...
+                                                <table class="divider_content" border="0" cellpadding="0" cellspacing="0" width="100%" style="table-layout: fixed; vertical-align: top; border-spacing: 0; border-collapse: collapse; mso-table-lspace: 0; mso-table-rspace: 0; border-top: 0 solid #BBBBBB; width: 100%;" align="center" role="presentation" valign="top">
+                                                    <tbody>
+                                                    <tr style="vertical-align: top;" valign="top">
+                                                        <td style="word-break: break-word; vertical-align: top; -ms-text-size-adjust: 100%; -webkit-text-size-adjust: 100%;" valign="top"><span></span></td>
+                                                    </tr>
+                                                    </tbody>
+                                                </table>
+                                            </td>
+                                        </tr>a
```
huang0h, line 223: ⛏️ a (can we get rid of this extra a? it shows up in the template and it looks a little weird)

image

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
SurabhiKeesara, line 942: Do we want to notify people that their image was rejected/deleted even if a reason isn't provided?

huang0h (changes requested):
Looks great! Just a couple quick comments:

SurabhiKeesara (commented):
Looks great, just a few questions/clarifications

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
CerberusLatrans: yeah this is based on the s3requester legacy code, unsure of the reason behind 10

api/src/main/java/com/codeforcommunity/dto/emailer/EmailAttachment.java
CerberusLatrans: oops good catch

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
CerberusLatrans: I guess yes (https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/pull/222#discussion_r1485656681)

CerberusLatrans:
Now throws error if trying to reject an approved image (should never happen).
Now deletes and sends rejection email with default message if rejectionReason is null.
Screenshot 2024-03-24 at 3 31 35 PM

service/src/main/java/com/codeforcommunity/requester/Emailer.java
```diff
@@ -23,6 +23,8 @@ public class Emailer {
       PropertiesLoader.loadProperty("email_subject_account_deleted");
   private final String subjectEmailNeighborhoods =
       PropertiesLoader.loadProperty("email_subject_neighborhood_notification");
+  private final String subjectImageRejected =
+      PropertiesLoader.loadProperty("email_subject_image_rejected");
```
huang0h, line 27: If we want to load the email subject this way, we'll want to add this to `server.properties.example` - you can simply match what we do for the other email subjects. 

When the backend gets deployed to AWS, it'll copy the property name and value into the `server.properties` and be available here; if you're really interested you can take a look at `deploy.py` for more details

huang0h (changes requested):

huang0h:
This is out of scope for this ticket, but I think it'd be a good idea to attach the image that was rejected to the email so the user knows which of their images got rejected, once we finish implementing email attachments as well - this'll probably be another ticket.

service/src/main/java/com/codeforcommunity/requester/Emailer.java
CerberusLatrans: Thanks for the info, good catch! Ill add it

huang0h (approved):
lgtm, great work here!

PR #224, opened 2024-02-25
Sk Add Get Unapproved Images (Filter Site Images)

### LINKED ISSUE #218: Add Route to Get Filtered Site Images
**Why**
- in allowing admins to approve/reject site images, admins can approve or reject them on the admin page somewhere
- additionally, allow admins to find specific 

**What to do**
- add a route to grab all the unapproved images with the filters in `ProtectedSiteProcessorImpl.java`
- See [the API reference](https://docs.c4cneu.com/sftt/sftt-api-spec/#get-unapproved-images-admin-only) for more details
- Use the [filterSites method in the ProtectedSiteProcessor](https://github.com/Code-4-Community/speak-for-the-trees-backend-v2/blob/master/service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java#L715) as an implementation guide:
  - Dynamically generate the condition by starting with selecting unapproved images, and adding to it based on any non-null parameters
  - Join across the `site_images`, `site_entries`, `sites`, and `users` table to get the necessary data
  - Filter based on: `uploaded_at`, `image_url`, `anonymous`, `uploader_id`, `site_entry_id` and `id`

SurabhiKeesara:
## Why

Resolves #218 

This PR resolves the ability to get unapproved site images filtered by upload date, site ID, and neighborhood IDs

## This PR 

- Use existing unapproved_images route
- Detailed here: https://docs.c4cneu.com/sftt/sftt-api-spec/#get-unapproved-images-admin-only

## Verification Steps

- Upload images with https://docs.c4cneu.com/sftt/sftt-api-spec/#upload-site-image-admin-and-owner-only
- GET http://localhost:8081/api/v1/protected/sites/unapproved_images, specifying filter parameters
- Reference the sample postman request below. Note that all filter fields are specified as params, similar to filter sites

Screenshot 2024-02-25 at 12 29 00 PM

### DIFF
--- a/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
+++ b/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
@@ -6,6 +6,7 @@ import com.codeforcommunity.dto.site.AddSitesRequest;
 import com.codeforcommunity.dto.site.AdoptedSitesResponse;
 import com.codeforcommunity.dto.site.EditSiteRequest;
 import com.codeforcommunity.dto.site.EditStewardshipRequest;
+import com.codeforcommunity.dto.site.FilterSiteImageRequest;
 import com.codeforcommunity.dto.site.FilterSitesRequest;
 import com.codeforcommunity.dto.site.FilterSitesResponse;
 import com.codeforcommunity.dto.site.NameSiteEntryRequest;
@@ -105,6 +106,22 @@ public interface IProtectedSiteProcessor {
    * </ul>
    */
   List<FilterSitesResponse> filterSites(JWTData userData, FilterSitesRequest filterSitesRequest);
+
+  /**
+   * Retrieves unapproved site images based on site filters, as detailed below.
+   *
+   * <p>The optional criteria are:
+   *
+   * <ul>
+   *   <li>The species of the site's latest site entry is in `treeSpecies`
+   *   <li>The site has been adopted by a user on or after `adoptedStart`
+   *   <li>The site has been adopted by a user on or before `adoptedEnd`
+   *   <li>The site's latest stewardship activity was recorded on or after `lastActivityStart`
+   *   <li>The site's latest stewardship activity was recorded on or before `lastActivityEnd`
+   *   <li>The site is located in a neighborhood in `neighborhoodIds`
+   * </ul>
+   */
+  List<FilterSitesResponse> filterSiteImages(JWTData userData, FilterSiteImageRequest filterSiteImageRequest);
 
   /** Edits the site entry with the given entryId */
   void editSiteEntry(JWTData userData, int entryId, UpdateSiteRequest editSiteEntryRequest);
--- a/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
+++ b/api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
@@ -7,6 +7,7 @@ import com.codeforcommunity.dto.site.AdoptedSitesResponse;
 import com.codeforcommunity.dto.site.EditSiteRequest;
 import com.codeforcommunity.dto.site.EditStewardshipRequest;
 import com.codeforcommunity.dto.site.FilterSiteImageRequest;
+import com.codeforcommunity.dto.site.FilterSiteImageResponse;
 import com.codeforcommunity.dto.site.FilterSitesRequest;
 import com.codeforcommunity.dto.site.FilterSitesResponse;
 import com.codeforcommunity.dto.site.NameSiteEntryRequest;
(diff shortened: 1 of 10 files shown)


api/src/main/java/com/codeforcommunity/dto/site/FilterSiteImageResponse.java
```diff
@@ -0,0 +1,109 @@
...
+import java.sql.Date;
+
+public class FilterSiteImageResponse {
+
+  private int imageId;
+  private String imageUrl;
+  private int siteId;
+  private String uploaderName;
+  private String uploaderEmail;
+  private String dateSubmitted;
```
huang0h, line 12: ⛏️ Can we convert the `dateSubmitted` field to a regular java `Date`, then add this decorator above it: `@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "MM/dd/yyyy")`
This will automatically convert the object into a string in the response body and let us easily change the format if we ever want to.

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -852,6 +855,86 @@ public List<FilterSitesResponse> filterSites(
...
+                .where(filterCondition)
+                .fetch();
+
+    return imagesFiltered.stream()
+        .map(
+            rec -> {
+              String uploaderName = rec.get(USERS.FIRST_NAME) + ' ' + rec.get(USERS.LAST_NAME);
+
+              String dateSubmitted =
+                  rec.get(SITE_IMAGES.UPLOADED_AT).toLocalDateTime().toLocalDate().toString();
```
huang0h, line 922: ⛏️ Related to the previous comment - instead of processing the upload time here, we can just directly pass the record's field into the response object, and it'll automatically get formatted in the response.

huang0h (changes requested):
lgtm! Just one quick comment:

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
```diff
@@ -420,7 +424,48 @@ private void handleApproveSiteImage(RoutingContext ctx) {
...
+        RestFunctions.getOptionalQueryParam(ctx, "submittedEnd", (date) -> new Timestamp(Date.valueOf(date).getTime()));
+    Optional<List<Integer>> siteIds =
+        RestFunctions.getOptionalQueryParam(
+            ctx,
+            "siteIds",
+            (string) ->
+                Arrays.stream(string.split(","))
+                    .map(Integer::parseInt)
+                    .collect(Collectors.toList()));
+    Optional<List<Integer>> neighborhoodIds =
```
CerberusLatrans, line 445: Optional since this might be the only case of it, but do you think it would make sense to abstract this repeated code out into a helper? I feel like parsing a list of optional query parameters would be quite a common pattern in the codebase  but I might be wrong. In the case that it is, we might be able to use generics for the parsing.

api/src/main/java/com/codeforcommunity/api/IProtectedSiteProcessor.java
```diff
@@ -106,6 +108,21 @@ void uploadSiteImage(
    */
   List<FilterSitesResponse> filterSites(JWTData userData, FilterSitesRequest filterSitesRequest);
 
+  /**
+   * Retrieves unapproved site images, as detailed below
```
CerberusLatrans, line 112: Maybe we can rename the method to `filterUnapprovedSiteImages` so that in the future someone might not think that it is filtering from the set of all site images?

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
```diff
@@ -852,6 +855,86 @@ public List<FilterSitesResponse> filterSites(
         .collect(Collectors.toList());
   }
 
+  @Override
+  public List<FilterSiteImageResponse> filterSiteImages(
+      JWTData userData, FilterSiteImageRequest filterSiteImageRequest) {
+    assertAdminOrSuperAdmin(userData.getPrivilegeLevel());
+
+    Condition filterCondition =
+        SITE_IMAGES.APPROVAL_STATUS.eq(String.valueOf(ImageApprovalStatus.SUBMITTED));
```
CerberusLatrans, line 864: I might be looking at the wrong spot but it seems like you aren't reusing the getUnapprovedImages route. However, I think this is fine since the logic would get pretty messy.

SurabhiKeesara: This is a good point and I did look at getUnapprovedImages. However, I think the filtering functionality makes it a little bit messy. If you have more to say on what you're thinking with getUnapprovedImages though I can revisit this

huang0h: I think `getUnapprovedImages` was from the very start of building the site images stuff, when we hadn't yet decided to add filtering on anything. I don't think we'll have a reason to use this route anymore unfortunately so we'll probably want to remove it at some point

huang0h (approved):
lgtm 👍

PR #230, opened 2024-05-07
Ah.site reporting edit

huang0h:
## Why

Resolves #<ticket number here>

Minor tweak to site reporting route - makes the description optional, and uses a default description if none is given
Main PR: https://github.com/Code-4-Community/speak-for-the-trees-frontend/pull/355

## Verification Steps

Verified description gets replaced with default if not given:

image


CerberusLatrans (approved):
looks fab

PR #232, opened 2024-05-31
Ot.rejected image

### LINKED ISSUE #228: Emailer: Add the rejected image to the reject image email
**Why**

Users should be able to know which of their images were rejected - currently, when we notify them of a rejection, we don't include the actual rejected email.

**What to do**

- Attach the rejected image to the rejection email via a `GetObjectRequest` (see `S3Requester.loadHTML` as an example of fetching an object from S3)
  - We can get an input stream via the `getObjectContent()` ([reference](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/S3Object.html#getObjectContent--)) on the object
  - We can get its MIME type by calling `getObjectMetadata().getContentType()` ([reference](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ObjectMetadata.html#getContentType--)) on the object

CerberusLatrans:
## Why

Resolves #228 

Attaches the rejected image to the rejection email sent.

## This PR
Directly passes the s3objectInputStream to a ByteArrayDataSource to construct an AttachmentResource.

Abstracts some email construction in EmailOperations.java for single and multiple recipient emails into helpers to allow for optional email attachments.

Also makes a rejection reason optional after #231 

## Verification Steps
Ensured that rejecting both jpeg and png image correctly attaches in the email

### DIFF
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
@@ -22,6 +22,7 @@ import com.codeforcommunity.dto.site.ReportSiteRequest;
 import com.codeforcommunity.dto.site.SiteEntryImage;
 import com.codeforcommunity.dto.site.UpdateSiteRequest;
 import com.codeforcommunity.dto.site.UploadSiteImageRequest;
+import com.codeforcommunity.exceptions.MissingParameterException;
 import com.codeforcommunity.rest.IRouter;
 import com.codeforcommunity.rest.RestFunctions;
 import io.vertx.core.Vertx;
@@ -416,7 +417,13 @@ public class ProtectedSiteRouter implements IRouter {
   private void handleRejectSiteImage(RoutingContext ctx) {
     JWTData userData = ctx.get("jwt_data");
     int imageId = RestFunctions.getRequestParameterAsInt(ctx.request(), "image_id");
-    String rejectionReason = RestFunctions.getRequestParameterAsString(ctx.request(), "reason");
+
+    String rejectionReason;
+    try {
+      rejectionReason = RestFunctions.getRequestParameterAsString(ctx.request(), "reason");
+    } catch (MissingParameterException e) {
+      rejectionReason = "Your image upload was rejected by an admin";
+    }
 
     processor.rejectSiteImage(userData, imageId, rejectionReason);
 
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
@@ -37,6 +37,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class ProtectedSiteRouter implements IRouter {
@@ -417,15 +418,11 @@ public class ProtectedSiteRouter implements IRouter {
   private void handleRejectSiteImage(RoutingContext ctx) {
     JWTData userData = ctx.get("jwt_data");
     int imageId = RestFunctions.getRequestParameterAsInt(ctx.request(), "image_id");
+    String defaultRejectionReason = "Your image upload was rejected by an admin";
+    Optional<String> rejectionReason = RestFunctions.getOptionalQueryParam(
+            ctx, "reason", Function.identity());
 
-    String rejectionReason;
-    try {
-      rejectionReason = RestFunctions.getRequestParameterAsString(ctx.request(), "reason");
-    } catch (MissingParameterException e) {
-      rejectionReason = "Your image upload was rejected by an admin";
-    }
-
-    processor.rejectSiteImage(userData, imageId, rejectionReason);
+    processor.rejectSiteImage(userData, imageId, rejectionReason.orElse(defaultRejectionReason));
 
     end(ctx.response(), 200);
   }
--- a/common/src/main/java/com/codeforcommunity/email/EmailOperations.java
+++ b/common/src/main/java/com/codeforcommunity/email/EmailOperations.java
@@ -4,6 +4,7 @@ import com.codeforcommunity.logger.SLogger;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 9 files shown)


api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
```diff
@@ -416,7 +417,13 @@ private void registerRejectSiteImage(Router router) {
...
     JWTData userData = ctx.get("jwt_data");
     int imageId = RestFunctions.getRequestParameterAsInt(ctx.request(), "image_id");
-    String rejectionReason = RestFunctions.getRequestParameterAsString(ctx.request(), "reason");
+
+    String rejectionReason;
+    try {
+      rejectionReason = RestFunctions.getRequestParameterAsString(ctx.request(), "reason");
+    } catch (MissingParameterException e) {
+      rejectionReason = "Your image upload was rejected by an admin";
+    }
```
huang0h, line 426: ⛏️ Instead of doing a try/catch, can we instead use `RestFunctions.getoptionalQueryParam()` and work with an `Optional` instead? It's generally better to avoid using exceptions as control flow if we can.

common/src/main/java/com/codeforcommunity/email/EmailOperations.java
```diff
@@ -150,6 +151,34 @@ private void sendEmail(Email email, String subject) {
...
+   */
+  public void sendEmailToOneRecipient(
+          String sendToName, String sendToEmail, String subject, String emailBody, List<AttachmentResource> attachments) {
+    if (!shouldSendEmails) {
+      return;
+    }
+    logger.info(String.format("Sending email with subject `%s`", subject));
+    Email email = buildEmailSingleRecipient(sendToName, sendToEmail, subject, emailBody, attachments);
+    this.sendEmail(email, subject);
+  }
```
huang0h, line 180: ⛏️ The helper method `buildEmailSingleRecipient` is a good idea, but is there any reason we can't use regular Java method overloading here and have the other version of `sendEmailToOneRecipient` just call this version and pass an empty list for `attachments`? IMO that would keep things simple and avoid code duplication.

common/src/main/java/com/codeforcommunity/email/EmailOperations.java
```diff
@@ -159,38 +188,44 @@ public void sendEmailToOneRecipient(
...
 
+  /** Send an email with the given subject and body to the users with the given email addresses with attachments. */
+  public void sendEmailToMultipleRecipients(
+      HashSet<String> sendToEmails, String subject, String emailBody, List<AttachmentResource> attachments) {
+    if (!shouldSendEmails) {
+      return;
+    }
+    logger.info(String.format("Sending emails with subject `%s`", subject));
+    Email email = buildEmailMultipleRecipient(sendToEmails, subject, emailBody, attachments);
```
huang0h, line 217: ⛏️ Same idea as the above comment - is there something stopping us from just using method overloading here?

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
```diff
@@ -233,6 +242,37 @@ public static void deleteSiteImage(String imageUrl) {
...
+    DataSource datasource;
+    try {
+      datasource = new ByteArrayDataSource(is, mimeType);
+    } catch (IOException e) {
+      throw new MalformedParameterException("Image encoding is incompatible");
+    }
+
+    String name = image.getKey();
+    return new AttachmentResource(name, datasource);
+  }
```
huang0h, line 274: nit: IMO I think this part of the method should live somewhere else (probably in the `loadSiteImage()` you added here), and have this method just return the S3object we fetched. The `S3Requester` class should be concerned only about doing S3 stuff, and converting an image into an emailer-specific attachment class should fall under the `Emailer`'s concerns. 

This separation of concerns keeps our classes and methods logically organized and reusable for the future - if we ever wanted to fetch an object from S3 and do something else besides turn it into an `AttachmentResource`, we'll already have a nice method to use,

service/target 2/surefire-reports/com.codeforcommunity.processor.ReservationProcessorImplTest.txt
```diff
@@ -0,0 +1,4 @@
+-------------------------------------------------------------------------------
+Test set: com.codeforcommunity.processor.ReservationProcessorImplTest
+-------------------------------------------------------------------------------
+Tests run: 20, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.673 s - in com.codeforcommunity.processor.ReservationProcessorImplTest
```
huang0h, line 4: btw if you want you can remove this file and anything in `node_modules` - honestly this should be in the `.gitignore`, not sure why it isn't

huang0h (commented):
Great job! Overall approach looks good, just a couple comments about code organization:

service/src/main/java/com/codeforcommunity/requester/S3Requester.java
CerberusLatrans: I agree, something felt a little wrong about the organization so thanks for spelling it out.

common/src/main/java/com/codeforcommunity/email/EmailOperations.java
CerberusLatrans: I was considering but was afraid to break the existing non-attachment email functionality. Will refactor

CerberusLatrans:
refactored all points of interest. retested with reason and missing reason

common/src/main/java/com/codeforcommunity/email/EmailOperations.java
```diff
@@ -179,20 +179,6 @@ public void sendEmailToOneRecipient(
...
-  public void sendEmailToOneRecipient(
-      String sendToName, String sendToEmail, String subject, String emailBody) {
-    if (!shouldSendEmails) {
-      return;
-    }
-    logger.info(String.format("Sending email with subject `%s`", subject));
-    Email email = buildEmailSingleRecipient(sendToName, sendToEmail, subject, emailBody, new ArrayList<>());
-    this.sendEmail(email, subject);
-  }
-
```
huang0h, line 164: Up to you if you want to make any more changes here - for future reference, we should be able to do something like this method while avoiding duplicating code by just having its body call the other version with an empty list, so something like this:

```
public void sendEmailToOneRecipient(String sendToName, String sendToEmail, String subject, String emailBody) {
  this.sendEmailToOneRecipient(sendToName, sendToEmail, subject, emailBody, new ArrayList<>());
}
```

huang0h (approved):
lgtm! 💯

PR #233, opened 2024-06-02
Implement Site Entry Deleting

### LINKED ISSUE #211: Implement route: delete site entry
**Why**
- Admins should be able to delete individual entries for a site to remove outdated/obsolete entries

**What to do**
- See [the route's API spec ](https://docs.c4cneu.com/sftt/sftt-api-spec/#delete-site-entry-admin-only) for implementation details
- All sites must have at least one site entry - if we try to delete the only entry for a given site, it should instead return an error
- When we delete an entry, we also need to delete any other information associated with that entry
    - Tables `entry_usernames` and `site_images` are associated with site entries - rows in these tables should be deleted too
    - Images also need to deleted from AWS - we can use `S3Requester.deleteSiteImage()` to do so

huang0h:
## Why

Resolves #211 

Allow admins to delete site entries, such as if they are old/obsolete

## This PR

- Adds the `deleted_at` column to the `site_entries` table
- Adds a route to soft-delete entries by setting the `deleted_at` column
- Update the get site route to only fetch non-deleted entries

## Verification Steps

- Verified that deleting the route set the `deleted_at` field
- Verified that the route returns an error if there are no other entries
- Verified on the site that deleted entries don't show up

### DIFF
--- a/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
@@ -17,7 +17,6 @@ import com.codeforcommunity.dto.site.NameSiteEntryRequest;
 import com.codeforcommunity.dto.site.ParentAdoptSiteRequest;
 import com.codeforcommunity.dto.site.ParentRecordStewardshipRequest;
 import com.codeforcommunity.dto.site.RecordStewardshipRequest;
-import com.codeforcommunity.dto.site.RejectImageRequest;
 import com.codeforcommunity.dto.site.ReportSiteRequest;
 import com.codeforcommunity.dto.site.SiteEntryImage;
 import com.codeforcommunity.dto.site.UpdateSiteRequest;
@@ -73,6 +72,7 @@ public class ProtectedSiteRouter implements IRouter {
     registerApproveSiteImage(router);
     registerGetUnapprovedImages(router);
     registerReportSiteIssue(router);
+    registerDeleteSiteEntry(router);
     return router;
   }
 
@@ -364,9 +364,7 @@ public class ProtectedSiteRouter implements IRouter {
         RestFunctions.getOptionalQueryParam(ctx, "lastActivityEnd", Date::valueOf);
     Optional<List<Integer>> neighborhoodIds =
         RestFunctions.getOptionalQueryParam(
-            ctx,
-            "neighborhoodIds",
-            this::parseOptionalQueryParamList);
+            ctx, "neighborhoodIds", this::parseOptionalQueryParamList);
     Optional<Integer> activityCountMax =
         RestFunctions.getOptionalQueryParam(ctx, "activityCountMax", Integer::parseInt);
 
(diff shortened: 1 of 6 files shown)


service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
```diff
@@ -293,7 +294,7 @@ public SiteGeoResponse getSiteGeoJson() {
   private static class SiteGeoResponseCache {
     private static SiteGeoResponse response;
     private static long expireTime = 0;
-    private static final long timeToLive = 10000; // in milliseconds, 5000 is the avg. response time
+    private static final long timeToLive = 1 * 60 * 1000; // in milliseconds, 5000 is the avg. response time
```
huang0h, line 297: Upping the cache time to 1 minute - we probably could do more, but it also means deleting/adding sites won't be visible until after the cache expires

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
```diff
@@ -446,19 +444,16 @@ private void handleFilterUnapprovedSiteImages(RoutingContext ctx) {
...
     Optional<Timestamp> submittedEnd =
-        RestFunctions.getOptionalQueryParam(ctx, "submittedEnd", (date) -> new Timestamp(Date.valueOf(date).getTime()));
-    Optional<List<Integer>> siteIds =
         RestFunctions.getOptionalQueryParam(
-            ctx,
-            "siteIds",
-            this::parseOptionalQueryParamList);
+            ctx, "submittedEnd", (date) -> new Timestamp(Date.valueOf(date).getTime()));
+    Optional<List<Integer>> siteIds =
+        RestFunctions.getOptionalQueryParam(ctx, "siteIds", this::parseOptionalQueryParamList);
```
CerberusLatrans, line 453: Just to make sure, this is all formatting stuff an no logic change

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
CerberusLatrans: this is to cache the data for the home page map? What is the reasoning behind upping cache time to minute?

CerberusLatrans: is since filtering takes extra query time and so less frequent cache rates will not make it so expensive

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
CerberusLatrans, line 1: Screenshot 2024-06-12 at 10 22 31 PM
are we discarding this part of the spec since we are now only soft deleting?

CerberusLatrans:
Nice job. One more thing-- if we are soft deleting, we will also probably want to account for that in the `checkEntryExists` of ProtectedSiteProcessorImpl.java

In that same file, 'latestSiteEntry` also should prob do the filter.

CerberusLatrans:
Also probably `fiterSites` in the same file. Sorry this is annoying but these two methods in `SiteProcessorImpl.java` might also need the filter?

Screenshot 2024-06-12 at 10 37 36 PM

service/src/main/java/com/codeforcommunity/processor/MapProcessorImpl.java
huang0h: Yeah, since this method requires a lot of work to query + process the data, the idea here is that increasing the cache time will reduce the load on our database/backend. Since it's just a super simple cache though, it'll mean any updates to the map may take up to a minute to show up for the users - I think I'm gonna dial this back just a bit so the UX doesn't suffer too much, and if it does get really bad we can start re-thinking how we cache the response.

api/src/main/java/com/codeforcommunity/rest/subrouter/ProtectedSiteRouter.java
huang0h: I think this is all from the auto-formatter, yeah

service/src/main/java/com/codeforcommunity/processor/ProtectedSiteProcessorImpl.java
huang0h: Yeah, I wrote that part a while ago and it needs updating (along with a whole bunch of other routes 😞)

huang0h:
> Nice job. One more thing-- if we are soft deleting, we will also probably want to account for that in the `checkEntryExists` of ProtectedSiteProcessorImpl.java
> 
> In that same file, `latestSiteEntry` also should prob do the filter.

>Also probably fiterSites in the same file. Sorry this is annoying but these two methods in SiteProcessorImpl.java might also need the filter?

Thanks for calling these out - missed these, we'll definitely want to check for deletion as well. `calculateBenefits` should probably just use `latestSiteEntry` as well.

CerberusLatrans (approved):
Looks gucci to me!

PR #237, opened 2024-07-19
Ot.rate limit v2 (Bucket4j server side throttling)

CerberusLatrans:
## Why

Resolves #<ticket number here>

Implements server-side IP-specific token bucket API throttling to prevent server overload from massive popularity or DOS attack.

## This PR

IpThrottlingFilter class instance acts as a VertX handler and intercepts all API calls before they get forwarded to processors. Keeps a hash map of IP addresses to buckets. Adds new buckets if new IP address. Uses bucket4j library.

Currently there is only one throttling class instance, so it is a globally-aggregated token consumption per IP. If we wanted to throttle per router or per route, we would create separate filter class instances for each route(r).

These are subject to change, but initial capacity is 20 and refills tokens at 10/min greedily (1 token every 6 sec) . One API route call consumes one token.

Service main contains these constants, creates an IpThrottlingFilter instance with the config (with the constants) and passes it to ApiRouter. The instance gets mounted to the common and protected router. If we wanted to to lower route(r) level throttling, we would create individual instances for each route(r) during subrouter initialization or instantiation.

## Verification Steps

Postman spamming to ensure that you get 429 error. Localhost reload spamming to get 429 error. Frontend might want to display different message when 429 occurs (Too many requests)?

Screenshot 2024-07-19 at 4 39 05 PM
Screenshot 2024-07-19 at 4 39 56 PM

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -26,6 +26,12 @@
             <groupId>io.vertx</groupId>
             <artifactId>vertx-web</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.bucket4j</groupId>
+            <artifactId>bucket4j-core</artifactId>
+            <version>8.0.1</version>
+        </dependency>
     </dependencies>
 
 </project>
--- a/api/src/main/java/com/codeforcommunity/rest/ApiRouter.java
+++ b/api/src/main/java/com/codeforcommunity/rest/ApiRouter.java
@@ -47,6 +47,7 @@ public class ApiRouter implements IRouter {
   private final ReportRouter reportRouter;
   private final ProtectedNeighborhoodsRouter protectedNeighborhoodsRouter;
   private final ProtectedEmailerRouter protectedEmailerRouter;
+  private final IpThrottlingFilter filter;
 
   public ApiRouter(
       IAuthProcessor authProcessor,
@@ -62,7 +63,8 @@ public class ApiRouter implements IRouter {
       IReportProcessor reportProcessor,
       IProtectedNeighborhoodsProcessor protectedNeighborhoodsProcessor,
       IProtectedEmailerProcessor emailerProcessor,
-      JWTAuthorizer jwtAuthorizer) {
+      JWTAuthorizer jwtAuthorizer,
+      IpThrottlingFilter filter) {
     this.commonRouter = new CommonRouter(jwtAuthorizer);
     this.authRouter = new AuthRouter(authProcessor);
     this.protectedUserRouter = new ProtectedUserRouter(protectedUserProcessor);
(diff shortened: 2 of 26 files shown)


CerberusLatrans:
Should have cleaned up PR to exclude the non-relevant changes from pulling from master but I will just leave a comment under the changed files to make it easier

api/pom.xml
CerberusLatrans, line 1: changed

service/src/main/java/com/codeforcommunity/ServiceMain.java
CerberusLatrans, line 1: changed

api/src/main/java/com/codeforcommunity/rest/ApiRouter.java
CerberusLatrans, line 1: changed

api/src/main/java/com/codeforcommunity/rest/IpThrottlingFilter.java
CerberusLatrans, line 1: changed

api/src/main/java/com/codeforcommunity/rest/IpThrottlingFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  public void handle(RoutingContext ctx) {
+    String ip = ctx.request().remoteAddress().toString();
+    Bucket bucket = this.buckets.get(ip);
+
+    if (bucket == null) {
+      bucket = Bucket.builder().addLimit(this.config.getBandwidths()[0]).build();
+      this.buckets.put(ip, bucket);
+    }
```
huang0h, line 33: 💭 Is there a config option to expire and delete buckets after a period of inactivity? There's a potential to log so many unique IPs that the bucket map consumes all the available memory. I don't think it's likely since that'd require a huge amount of unique IPs to do so - if it's not a lot of work, feel free to implement it, otherwise don

service/src/main/java/com/codeforcommunity/ServiceMain.java
```diff
@@ -33,13 +33,20 @@
...
 import io.vertx.core.Vertx;
+import java.time.Duration;
 import java.util.Properties;
 import org.jooq.DSLContext;
 import org.jooq.impl.DSL;
 
 public class ServiceMain {
   private DSLContext db;
+  public static final long CAPACITY = 20;
+  public static final Refill REFILL = Refill.greedy(10, Duration.ofMinutes(1));
```
huang0h, line 49: ❓ How'd you determine the initial capacity and refill rate - are these some common numbers or just gut feelings? 

IMO we might want to bump up either the capacity or the refill rate, since there are some actions that might occur in quick succession such as approving/rejecting/uploading images which are one request per image (tbh it might be a good idea to switch to bulk routes here, even though it was my idea to do it one at a time)

huang0h (commented):
Overall seems to make sense. This wasn't an easy ask, so thanks for looking into this and taking this on! A couple questions before fully approving:

huang0h:
> Should have cleaned up PR to exclude the non-relevant changes from pulling from master but I will just leave a comment under the changed files to make it easier

FYI for the future, you can use `git status` in the terminal to check for files that are going to be added/committed, and then you can find any files that you didn't mean to edit. Always to make sure you know what's been changed before you commit!

service/src/main/java/com/codeforcommunity/ServiceMain.java
CerberusLatrans: These are arbitrary gut feeling values. wdym by bulk routes? do you mean router specific instead of global?

api/src/main/java/com/codeforcommunity/rest/IpThrottlingFilter.java
CerberusLatrans: good idea. will do.

huang0h (approved):
💯 💯 💯 💯 💯 💯 💯

